PR #3, opened 2015-07-20
Delete .bowerrc

azat-co:
as @jdamick proposed, let's delete this.


amitmawkin:
It should regenerate the default
. bowerrc on a build

jdamick:
For people just trying to check this out and use it the .bowerrc won't be generated..

amitmawkin:
Yeah so we hAve not deleted it but remove the proxy section,

Thanks




AM
=======
804-229-3241

Sent with Good (www.good.com)


-----Original Message-----

PR #5, opened 2015-07-22
Fixed low-level licensing, readme property descriptors, simplified property files, and removed core snapshot dependency tags in poms for feature collectors

amp5208:


amp5208:
I noticed that the index.html file was missing for the UI, causing the issue noted in the repo by a user.  Adding this file back in should fix this.

jdamick:
reviewed - looks good, thanks

PR #11, opened 2015-07-23
Fixed jira transient dependency issue

amp5208:
updated jira dependencies to match that of contents in HEAD of artifact prior to official public release (fixed jira transient dependency issue)

### DIFF
--- a/JiraFeatureCollector/pom.xml
+++ b/JiraFeatureCollector/pom.xml
@@ -5,7 +5,7 @@
 	<groupId>com.capitalone.dashboard</groupId>
 	<artifactId>jira-feature-collector</artifactId>
 	<packaging>jar</packaging>
-	<version>01.00.00.01</version>
+	<version>01.00.01.01</version>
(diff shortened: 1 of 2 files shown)


amp5208:
The dependency structure was re-tested using the following commands to ensure correct mapping, and test cases were run:  mvn clean compile test package.  This change should have been in the original, final merge to public github

JiraFeatureCollector/pom.xml
```diff
@@ -5,7 +5,7 @@
 	<groupId>com.capitalone.dashboard</groupId>
 	<artifactId>jira-feature-collector</artifactId>
 	<packaging>jar</packaging>
-	<version>01.00.00.01</version>
+	<version>01.00.01.01</version>
```
jdamick: this version scheme doesn't match the other plugins - can you update it to be semver style: x.x.x

amp5208:
Feature collector pom version numbers have been updated to match the syntax of other collectors - feel free to merge this pull request

PR #22, opened 2015-07-29
Corrected logic to create environment-component correctly.

tabladrum:
Moved the UDeployEnvResCompData object creation in the inner loop to create correct number of env-component objects.


amitmawkin:
merged for udeploy collector.

PR #29, opened 2015-08-03
Null pointer check to handle some badly formed json response.

tabladrum:
Found the issue with some empty deploy jobs in uDeploy.


amitmawkin:
Thanks for the changes Topo

PR #32, opened 2015-08-04
NPE check for empty dashboard components.

tabladrum:
For empty dashboard components, a NPE is thrown. Modified code to handle that situation.


tabladrum:
Mergred

PR #33, opened 2015-08-04
compress the image slightly so the page loads faster

jdamick:
the current image was > 2MB and is slow to load on github, now it's down to 450K, so it loads a bit faster.


amitmawkin:
Thanks I have merged the request

PR #34, opened 2015-08-05
Support for github.com. Still will not support access behind proxy.

tabladrum:
Please review the changes.


jdamick:
needs unit tests please

PR #43, opened 2015-08-07
maven to gradle

jdamick:
along with embedded mongo to enable testing
added sample properties file for clarity
updated README, for "hudson" as it is now looking for "Jenkins" properties

TODO: normalize names of directories to match and avoid confusion


msabramo:
has conflicts

PR #50, opened 2015-08-08
Fixed another bug with SourceControl widget

amitmawkin:
fixed another bug, since the scm field is introduced in widgetConfig options to track the type of SCM (Currently Github/Subversion) all the existing dashboard needs to resave the properties of the Source Code widget.


jdamick:
Really need unit tests for these...

PR #63, opened 2015-08-13
fixed .travis.yml

amitmawkin:


jdamick:
if we are switching to gradle then probably can hold off on the pom.xml change?

PR #79, opened 2015-08-14
Rename BuildCollector to HudsonCollector

msabramo:
because it is specific to Hudson/Jenkins.

Cc: @sudarkoff, @aconrad


jdamick:
agreed, but we need to choose either hudson or jenkins - since they are both used in multiple places (like the properties file)

msabramo:
Yeah, actually I would prefer Jenkins over Hudson. My plan (which I didn't voice) was to submit this PR to change `BuildCollector` to `HudsonCollector` (a relatively small change) and then after merge, submit another PR to rename `Hudson` to `Jenkins`.

msabramo:
Rebased to fix merge conflict.

msabramo:
Rebased again to fix merge conflict (caused by merge of my own change to `Setup.md`).

jdamick:
i think we should just leap frog and just rename to jenkins rather than hudson then jenkins like walked about on gitter..

amitmawkin:
the collector is now jenkins-build-collector, hence closing this pull request

PR #81, opened 2015-08-14
Add more debug logging to JiraFeatureCollector

msabramo:
This was stuff that I added while trying to figure out why the `JiraFeatureCollector` wasn't working for me last night.

Cc: @sudarkoff, @aconrad


tcrall:
More debug logging in the JiraFeatureCollector would be very nice.  I'm trying to configure it myself and getting Forbidden errors - it would be nice to see exactly what api call was being attempted so I could try to replicate it from the command line, etc.

PR #99, opened 2015-08-27
Make api module to run as a standalone (with the help of spring boot)

srid99:
Moved away from war to jar and must be able to run api in standalone mode just like the other collectors.
Updated the documentation and tested.


amitmawkin:
Sri  I evaluated it , I see you have introduced a new property file that contains the context root and port. Can you merge the properties as part of Environment properties under DASHBOARD_PROP or vice versa which ever is easier , end goal is to have one property file.

srid99:
Thought to have a separate property file just for SpringBoot. But it makes sense to have just one property file for convenience. Will do it.

srid99:
Hello there, Could you guys look into this PR ?

amitmawkin:
ok that fixed it , so I have merged it now

srid99:
Cool :+1:

PR #115, opened 2015-09-03
Docker containers for Hygieia and UI

pjaol:
Ability to run separate Docker containers for the API and the UI, similar to production layouts.

The API is using a tomcat container and the UI is using an nginx container
The docker images are part of the maven compile using spotify's docker plugin

```bash
cd api
mvn clean package docker:build

cd ../UI
mvn clean package docker:build
```
This will create 2 containers for you, hygieia_api and hygieia_ui
You can now create a swarm, connect a docker or managed mongo, to run it all within a standard docker-machine

```bash
docker run -h mongo -d --name mongo mongo:latest
docker run -h hygieia_api --link mongo:mongo -P -d --name hygieia_api hygieia_api
docker run -d -P -h hygieia_ui --name hygieia_ui --link hygieia_api:api hygieia_ui
```
hygieia_api expects a network link / alias called mongo
hygieia_ui expects a network link / alias called hygieia_api

PR #105, opened 2015-08-28
Docker support for Hygieia

tyrion70:
A simple setup to run Hygieia in a Docker container

PR #129, opened 2015-09-24
Added docker image creation support and documentation for API & UI

jdamick:
Upgraded dependencies & authentication to work with mongo 3 - also removed duplicate dependencies

Updated the pom to be able to build the docker image from the top level

Added docker-compose & setup docs

Fixed dependency issues - switched to use the spring-boot way, it was previously pulling in dependencies with different versions than spring-boot..


jdamick:
this pull addresses pull #105 & pull #115

jdamick:
@amitmawkin @tabladrum thoughts?

amitmawkin:
Thanks Jeff for contributing.

PR #137, opened 2015-10-05
Username was not unique

sidgoyal:
Looking at the Authentication Collection, it turns out that the usernames were not unique. Created a fix to have unique usernames, and also added a test for it. removed the findByUsernameAndPassword method in AuthenticationRepository because its redundant now with findByUsername and unique usernames

### DIFF
--- a/api/src/main/java/com/capitalone/dashboard/service/AuthenticationServiceImpl.java
+++ b/api/src/main/java/com/capitalone/dashboard/service/AuthenticationServiceImpl.java
@@ -1,7 +1,5 @@
 package com.capitalone.dashboard.service;
 
-import java.util.List;
-
 import org.bson.types.ObjectId;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
@@ -30,6 +28,7 @@ public class AuthenticationServiceImpl implements AuthenticationService {
     public Authentication get(ObjectId id) {
 
         Authentication authentication = authenticationRepository.findOne(id);
+        System.out.println("[DEBUG] GET authentication " + authentication);
         return authentication;
     }
 
@@ -72,11 +71,9 @@ public class AuthenticationServiceImpl implements AuthenticationService {
     @Override
     public boolean authenticate(String username, String password) {
         boolean flag = false;
-        List<Authentication> authenticationList = authenticationRepository.findByUsernameAndPassword(username, password);
-        if (authenticationList.size() == 1) {
-            if (authenticationList.get(0).getUsername().equals(username) && authenticationList.get(0).getPassword().equals(password)) {
-                flag = true;
-            }
+        Authentication authentication = authenticationRepository.findByUsername(username);
+        if (authentication.getUsername().equals(username) && authentication.getPassword().equals(password)) {
+        	flag = true;
         }
         return flag;
     }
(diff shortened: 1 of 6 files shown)


amitmawkin:
They are unique if you try to add a user that already exist you will get following error message 
screen shot 2015-10-05 at 6 30 08 pm

can you please check what version of code you are using ..

thanks,

sidgoyal:
I just recreated the error from the UI 

image


can you try running the dashboard with a new fresh mongoDB instance ? also, you can pull the test that I added and try running it, without the change in the model, it should fail.

common/src/test/java/com/capitalone/dashboard/repository/AuthenticationRepositoryTest.java
```diff
@@ -0,0 +1,60 @@
...
+    	try{
+    	Authentication user2 = new Authentication(username, "pass2");
+    	
+    	// This line should throw a DuplicateKeyException
+    	authenticationRepository.save(user2);
+    	
+    	// If the above line did not throw a DuplicateKeyException, fail the test
+    	fail("Didn't throw any Exception");
+    	}
+    	catch(DuplicateKeyException e){
```
jdamick: remove this catch, use: http://junit.sourceforge.net/javadoc/org/junit/Test.html#expected()

api/src/main/java/com/capitalone/dashboard/service/AuthenticationServiceImpl.java
```diff
@@ -72,11 +70,9 @@ public void delete(String username) {
     @Override
     public boolean authenticate(String username, String password) {
         boolean flag = false;
-        List<Authentication> authenticationList = authenticationRepository.findByUsernameAndPassword(username, password);
-        if (authenticationList.size() == 1) {
-            if (authenticationList.get(0).getUsername().equals(username) && authenticationList.get(0).getPassword().equals(password)) {
-                flag = true;
-            }
+        Authentication authentication = authenticationRepository.findByUsername(username);
+        if (authentication.getUsername().equals(username) && authentication.getPassword().equals(password)) {
```
jdamick: if you just looked it up by username why do you need to .equals it?

jdamick: can authentication be null?

jdamick:
either way, this is a good test that we should have.  thank you for identifying these issues.  it looks like we are comparing the clear-text password in here? @amitmawkin ?  that would be a major TODO for us...

amitmawkin:
I will investigate as I cannot produce this on my end

common/src/main/java/com/capitalone/dashboard/model/Authentication.java
```diff
@@ -11,8 +12,9 @@
 public class Authentication extends BaseModel {
 	
 	
-
+	@Indexed(unique = true)
```
jdamick: good catch

amitmawkin:
I checked on my end, i have an index turned on authentication collection.
so probably that explain why i am not facing this on my local mongodb collection
{
    "username" : 1,
    "password" : 1
}

common/src/test/java/com/capitalone/dashboard/repository/AuthenticationRepositoryTest.java
sidgoyal: Yeah, @Expected does not tell which line threw the exception, that's why I used the catch explicitly for the second save() call.

sidgoyal:
@jdamick  So I made the changes that you suggested. Also did some refactoring on the AuthenticationController to remove the DuplicateKeyException on get type methods and added null checks on AuthenticationSerivce

Issue #78, opened 2015-08-14
Passwords stored in clear


aconrad:
https://crackstation.net/hashing-security.htm

tabladrum:
The only place we have a password concept is the dashboard owner. Our initial version did not even have that - we believe all the information that are displayed on the dashboard should be transparent to everyone. The problem that we ran into was that PersonA could modify PersonB's dashboard. So, we introduced the owner concept just to make sure that does not happen.

Having said that, I already pushed some new API's and Encryption classes and am working on allowing real passwords (such as jenkins, jira passwords etc) to be stored encrypted.

msabramo:
Which passwords are we talking about and where are they stored?

tabladrum:
Not storing any yet other than "dashboard owner".

jdamick:
closing with pull

PR #142, opened 2015-10-14
hash user passwords

jdamick:
addresses issue #78 - this will now store user passwords hashed in sha512, which is not perfect, but better than plain text.


jdamick:
outstanding for a week - i'm merging

Issue #124, opened 2015-09-08
rename directories to match what they represent


jdamick:
Rather than generic names like "CodeQuality" we should rename the directories to meaning ones that represent what they are, like "SonarCollector" or something along those lines.

tabladrum:
@jdamick Close this?

PR #146, opened 2015-10-20
renamed poms & directories to match

jdamick:
there was a mixture of hyphen or camel-case, now they are all the same

issue #124


jdamick:
fyi @amitmawkin @tabladrum

amitmawkin:
I wil checkout this PR to see if it works correctly.

PR #148, opened 2015-10-22
Improve documentation

IonicaBizau:
Fixed grammar, and added/removes some commas. :dizzy:

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,25 +7,25 @@
 
 ### What is Hygieia
 
-#####Pronunciation:  	hi-gee-ya (origin Greek)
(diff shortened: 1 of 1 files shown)


jdamick:
we should revert, the grammar is not correct..

IonicaBizau:
@jdamick I will just remove the commas, would be that ok?

jdamick:
sounds good

IonicaBizau:
@jdamick Actually, can you please [leave some inline comments](https://github.com/capitalone/Hygieia/pull/148/files) what should be fixed? So, then I will fix it in another PR. :thought_balloon:

PR #152, opened 2015-10-26
fixes for PMD errors

jdamick:


jdamick:
light is green trap is clean.

Issue #162, opened 2015-11-04
Running mvn install package gives errors


joshku:
I did a fresh pull of the code from GitHub and tried to do the command `mvn clean install package` and I got the following output as a result:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] com.capitalone.dashboard:Hygieia ................... SUCCESS [  1.015 s]
[INFO] com.capitalone.dashboard:core ...................... FAILURE [ 11.820 s]
[INFO] com.capitalone.dashboard:api ....................... SKIPPED
[INFO] com.capitalone.dashboard:jenkins-build-collector ... SKIPPED
[INFO] com.capitalone.dashboard:sonar-codequality-collector SKIPPED
[INFO] com.capitalone.dashboard:subversion-collector ...... SKIPPED
[INFO] com.capitalone.dashboard:github-scm-collector ...... SKIPPED
[INFO] com.capitalone.dashboard:versionone-feature-collector SKIPPED
[INFO] com.capitalone.dashboard:udeploy-deployment-collector SKIPPED
[INFO] com.capitalone.dashboard:jira-feature-collector .... SKIPPED
[INFO] com.capitalone.dashboard:jenkins-cucumber-test-collector SKIPPED
[INFO] com.capitalone.dashboard:UI ........................ SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 13.855 s
[INFO] Finished at: 2015-11-04T13:52:38-05:00
[INFO] Final Memory: 41M/356M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.5:check (default) on project core: You have 21 PMD violations. For more details see: /home/ibmadmin/Hygieia/core/target/pmd.xml -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR]
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :core
```
[...]

PR #163, opened 2015-11-05
added exclusion for generated source & updated jacoco

jdamick:
Removed generated code
Added exclusion for generated code
Update jacoco


jdamick:
fix for issue #162

PR #164, opened 2015-11-05
Core support to handle fortify security scan.

tabladrum:
Core support to handle fortify security scan.


api/src/main/java/com/capitalone/dashboard/service/CodeQualityServiceImpl.java
```diff
@@ -1,61 +1,115 @@
...
+	public DataResponse<Iterable<CodeQuality>> search(CodeQualityRequest request) {
+		Component component = componentRepository.findOne(request
+				.getComponentId());
+		CollectorItem item = null;
+		Iterable<CodeQuality> result = null;
+		if (request.getType() == null) {
+			item = component.getCollectorItems().get(CollectorType.CodeQuality)
+					.get(0);
+		} else {
+			switch (request.getType()) {
```
jdamick: could optimize this and use an enum or minimize the duplicate code for better readability ..

api/src/main/java/com/capitalone/dashboard/service/CodeQualityServiceImpl.java
```diff
@@ -1,61 +1,115 @@
...
+		QCodeQuality quality = new QCodeQuality("quality");
+		BooleanBuilder builder = new BooleanBuilder();
+
+		builder.and(quality.collectorItemId.eq(item.getId()));
+
+		if (request.getNumberOfDays() != null) {
+			long endTimeTarget = new LocalDate()
+					.minusDays(request.getNumberOfDays()).toDate().getTime();
+			builder.and(quality.timestamp.goe(endTimeTarget));
+		} else {
```
jdamick: compact to else if

UI/src/components/widgets/codeanalysis/view.html
```diff
@@ -119,10 +119,20 @@
             </div>
             <div class="widget-body">
                 <table class="table">
-                    <tr class="dash-row"><td></td><td class="ca-name">Blocker</td><td class="ca-value">-</td></tr>
-                    <tr class="dash-row"><td></td><td class="ca-name">Critical</td><td class="ca-value">-</td></tr>
-                    <tr class="dash-row"><td></td><td class="ca-name">Major</td><td class="ca-value">-</td></tr>
-                    <tr class="dash-row"><td></td><td class="ca-name">Issues</td><td class="ca-value">-</td></tr>
```
jdamick: :+1:

UI/src/components/widgets/codeanalysis/view.js
```diff
@@ -72,6 +79,23 @@
...
         }
 
+        function processSaResponse(response) {
+            var deferred = $q.defer();
+            var saData = _.isEmpty(response.result) ? {} : response.result[0];
+
+            //ctrl.versionNumber = saData.version;
+
+            ctrl.securityIssues = [
+                getMetric(saData.metrics, 'blocker', 'Blocker'),
```
jdamick: do these not come from the api?

tabladrum:
@jdamick check the updates please. thanks.

jdamick:
so now in really looking at this, i think we should do the following:
- change to sub-resources for security and anything inside the quality widget, it's a bit kludgy to have a switch statement in the resource handler
  - have an actual sub-resource for security, so that it has its own path & mock data json
- this includes updating the test resource to be under the quality as well

We should think about having sub-widgets for quality, or break them into full widgets.  They don't follow the pattern right of the others so it's going to be harder to maintain / add to later on.  

Also the json returned / entered to the db should be normalized, it seems that field names like:
```ncloc``` & ```test_success_density``` are tool specific..

PR #189, opened 2015-11-25
Stash

dianewalls:
I've made a some modifications in order to collects data from a stash repo


tabladrum:
Thanks for the PR. We are reviewing it.

stash-scm-collector/.classpath
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jdamick: delete this file from the commit

jdamick:
please remove the dot files.

stash-scm-collector/.pmd
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jdamick: please remove

stash-scm-collector/.pmdruleset.xml
```diff
@@ -0,0 +1,1094 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jdamick: please remove file

stash-scm-collector/.project
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jdamick: please remove file

stash-scm-collector/.settings/org.eclipse.core.resources.prefs
```diff
@@ -0,0 +1,5 @@
+eclipse.preferences.version=1
```
jdamick: please remove file

stash-scm-collector/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,5 @@
+eclipse.preferences.version=1
```
jdamick: please remove file

jdamick:
Can you please add captures of request / responses from stash - so that we can have unit tests that verify the client.

dianewalls:
@jdamick, here is the json

https://gist.githubusercontent.com/anonymous/b06e40893a2457eed47f/raw/66a83193d4414d77f5c8df3cb4fc7ceaa8b006af/565cacebe4b01190df425508.json

I guess is the only needed. 

The query made by the client is like this.
https://mystashrepo.com/rest/api/1.0/projects/<project_key>/repos/<project_name>/commits?sha=master&since=2015-11-10T18:42Z

jdamick:
I still feel like there is insufficient test coverage for this plugin.. i'd like to take it but i'm not sure if it works. Could you please add a unit test that uses the captures of request / responses from stash.  I understand the other collectors have insufficient testing as well, and that's something that and needs to be addressed but we don't have a stash instance to verify against..

amitmawkin:
can we test against public bitbucket?  . Stash is now known as bitbucket 

https://www.atlassian.com/software/bitbucket/server

tcrall:
Stash has been rebranded as “Bitbucket Server” but Bitbucket server and “Bitbucket Cloud” do not have identical APIs.  I think you would need a different listener for each of them

dianewalls:
@amitmawkin a public bitbucket can be used. I created a default bitbucket client. It works 

you can test it with this repo url: https://bitbucket.org/dianewalls/vzevents.git and branch master.

This is my output 
output

And json is 
https://bitbucket.org/api/2.0/repositories/dianewalls/vzevents/commit/master/

The repo is public so it doesn't need authentication 

It can works with private but needs authentication. Maybe in this days I could make it work with a private.

The stash client and bitbucket client are almost the same. There are not many difference if you compare it.

srid99:
@jdamick @tabladrum As discussed in gitter I can write the unit test this weekend. Just to be clear, I will be writing unit test for classes GitCollectorTask, DefaultStashClient, DefaultBitbucketClient. The other classes doesn't seems to be interesting.

Also is it have to be unit test ? Can it also be integration test ? With a mock embedded server mocking Stash and BitBucket api's and using spring-boot test module we can test the collector api directly. And I can try my best to cover the logics. What do you guys think ? I am fine with both the ways but I would be happy if it can be integration test (makes my life easy) :)

jdamick:
@srid99 probably just terminology - here is example of a junit test:
https://github.com/square/okhttp/tree/master/mockwebserver
So it's something like:

```
public SomeTest {
   @Rule public MockWebServer server = new MockWebServer();
  @Test blah() {
....
  }
}
```

srid99:
So that's going to be integration test then :+1:

srid99:
I had a quick look at the PR and I have some questions, maybe @dianewalls can help me understanding it,

There are 2 GitClient implementations, one for BitBucket and another for Stash. The BitBucket client is annotated with @Component and Stash client is not. Based on my Spring knowledge (which is not that good :P), when Spring wires the dependencies for GitCollectorTask it will use BitBucket client and not the Stash client (since Spring will not recognize Stash as a component). Is that correct or am I missing something ?

dianewalls:
@srid99 you are right. I just remove the @component on stash client because I want to make the Bitbucket client work. I didn't want to make another collector because the only difference is the client. So maybe then we can have two threads

srid99:
@dianewalls I understand but I think we have to make a decision before we merge this PR (or not).

@amitmawkin @tabladrum @jdamick What is your opinion on this ? If we like to have separate collectors for each hosting service (github, stash, bitbucket and so on) then I would suggest to split (stash and bitbucket) into 2 collectors so that at least it is consistent even though they both are from same Company.

The other option would be to have one collector let's say git-scm-collector and provide a way to tell the collector which client should be used.

tabladrum:
Whats the difference between the two clients? Just some api endpoint patterns are different? Current Github collector handles both Enterprise and the Public even though the api urls are different. Can something similar be done?

srid99:
Had a quick comparison and Yes I there is a difference with the payload. Eg: in stash ithe scmRevisionNumber is mapped to id and bitbucket is is mapped to hash. To be precise the difference is with 3 fields, revision number, author and commit timestamp. So it needs a different client unless we use one client and make decision based on the field name eg: id or hash to decide if the response is from stash or bitbucket. But this sounds like a code smell to me :)

tabladrum:
I am voting for two distinct collectors in this case. Others: thoughts?

amitmawkin:
Just thinking out loud can a parent child relationship in spring boot take care of this based on type of repo
https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-spring-application.html#boot-features-fluent-builder-api
not sure if this will work, need to test it out.. thoughts

dianewalls:
I created a collector task for bitbucket and now I have both collector tasks running in the stash-scm-collector. But I'm not sure to commit my changes because you were voting.

srid99:
@amitmawkin Not sure if it fits for our need. I was looking for something like [registering the bean/implementation based on condition](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-developing-auto-configuration.html#boot-features-property-conditions). This way we can pass a property with value, say git.web.host.service.name=stash and every client should have this annotation with the expected value, eg: for Stash client it should be something like `@ConditionalOnProperty(prefix = "git.web.host.service.name", havingValue = "stash")`.

Also I see lots of code here is duplicated and also expect more duplication in future as we add more services (eg: gitlab and so). So it is worth to spend some time to see how we can reuse the most. In my opinion I only expect the implementation of clients to be different and the rest are going to be the same. So would suggest to have multi-module project ([explained in the previous comment](https://github.com/capitalone/Hygieia/pull/189#issuecomment-161412018)). This way still we can have separate collectors and reuse the common code.

But anyways this topic should not stop this PR. We can still merge it and the restructuring/refactor can be done in a separate PR.

jdamick:
i agree about the reuse and eliminate the duplicate code, but i think we should opt for something simpler rather a lot of spring annotations that makes the code hard to follow & maintain.

PR #230, opened 2015-12-17
Added capability for tests to run behind an authenticated corporate proxy

amp5208:
Current version of Hygieia does not allow for running tests (and thus building) via Maven from behind an authenticated proxy.  This update includes the ability for code to be fully built/tested/run from behind a corporate authentication proxy.  This will also allow for non-authenticated proxies to still be used as per normal.

Also, I updated the Setup readme to clearly indicate how a developer should configure a local proxy if they need to, given this enhancement.


jdamick:
The HTTP/S_PROXY variables should look like:

```
HTTP_PROXY=http://<id>:<passwd>@proxy.somewhere:80
```

So we shouldnt added HTTP_AUTHxxx, etc. as this breaks users..

softwareklinic:
Some of use auto proxy scripts… would that work as well?

Setup.md
```diff
@@ -79,6 +79,56 @@ You can pick and choose which collectors are applicable for your DevOps toolset
...
+Additionally, set the following export variables:
+
+```bash
+# Add these variables for authenticated proxy, required
+# for automated testing through an authenticated proxy:
+export HTTPAUTH_USER=companyId999
+export HTTPAUTH_PASS=yourPassword
+
+export HTTP_PROXY=http://proxy.kdc.capitalone.com:8099
+export HTTPS_PROXY=http://proxy.kdc.capitalone.com:8099
```
jdamick: should contain user creds info here..

core/src/test/java/com/capitalone/dashboard/repository/EmbeddedMongoDBRule.java
```diff
@@ -40,10 +42,28 @@
...
             String proxy = System.getenv("HTTP_PROXY");
+            final String authUser = System.getenv("HTTPAUTH_USER");
+            final String authPassword = System.getenv("HTTPAUTH_PASS");
+            
             if (proxy == null || proxy.isEmpty()) {
                 proxy = System.getProperty("HTTP_PROXY");
             }
             try {
+            	// Case for Proxy authentication required
+                if ((proxy != null && !proxy.isEmpty()) && (authUser != null && !authUser.isEmpty()) && (authUser != null && !authPassword.isEmpty())) {
```
jdamick: should pull the user & passwd from the URL.getUserInfo()

jdamick:
Can you give an example of an "auto proxy script"?

amp5208:
Are you referring to a proxy host running a .pac file?  if so, your .pac file should still run given the existing proxy logic, as the httpauth-user/pass variables will not even be added to the default authentication set by java.net as long as those system variables are not set.

Please confirm if this works for your use case.

amp5208:
Closing pull request and re-submitting with modified content.

PR #233, opened 2015-12-17
Enabled authenticated proxy configuration for embedded MongoDB tests

amp5208:
This enhancement enables the use and running of tests leveraging the embedded MongoDB test packages when an authenticated proxy is required.  This will also work with existing, non-authenticated proxy settings, as the logic for that option will run independent of the logic for authenticated proxies.

I also updated the Setup.md readme to include explicit instructions for how to configure both an authenticated and non-authenticated proxy.

### DIFF
--- a/Setup.md
+++ b/Setup.md
@@ -110,17 +110,12 @@ Update your Maven settings.xml file:
 Additionally, set the following export variables:
 
 ```bash
-# Add these variables for authenticated proxy, required
-# for automated testing through an authenticated proxy:
-export HTTPAUTH_USER=companyId999
-export HTTPAUTH_PASS=yourPassword
-
-export HTTP_PROXY=http://your.proxy.domain.name:8080
-export HTTPS_PROXY=http://your.proxy.domain.name:8080
-export JAVA_OPTS="$JAVA_OPTS -Dhttp.proxyHost=your.proxy.domain.name -Dhttp.proxyPort=8080 -Dhttp.proxyUser=$HTTPAUTH_USER -Dhttp.proxyPassword=$HTTPAUTH_PASS"
+export HTTP_PROXY=http://companyId999:yourPassword@your.proxy.domain.name:8080
+export HTTPS_PROXY=http://companyId999:yourPassword@your.proxy.domain.name:8080
+export JAVA_OPTS="$JAVA_OPTS -Dhttp.proxyHost=your.proxy.domain.name -Dhttp.proxyPort=8080 -Dhttp.proxyUser=companyId999 -Dhttp.proxyPassword=yourPassword"
 # This option may be duplicative if you have already updated your
 # Maven settings.xml file, but will only help:
-export MAVEN_OPTS="$MAVEN_OPTS -Dhttp.proxyHost=your.proxy.domain.name -Dhttp.proxyPort=8080 -Dhttp.proxyUser=$HTTPAUTH_USER -Dhttp.proxyPassword=$HTTPAUTH_PASS"
+export MAVEN_OPTS="$MAVEN_OPTS -Dhttp.proxyHost=your.proxy.domain.name -Dhttp.proxyPort=8080 -Dhttp.proxyUser=companyId999 -Dhttp.proxyPassword=yourPassword"
 ```
 
 Tests should now run/pass when built from behind a corporate proxy, even if it is an authenticated proxy
--- a/Setup.md
+++ b/Setup.md
@@ -27,13 +27,13 @@ The following components are required to run Hygieia:
          > use dashboardb
          switched to db dashboardb
          > db.createUser(
-              ... {
-                ... user: "dashboarduser",
-                ... pwd: "1qazxSw2",
-                ... roles: [
-                  ... {role: "readWrite", db: "dashboarddb"}
-                        ... ]
-                ... })
+                  {
+                    user: "dashboarduser",
+                    pwd: "1qazxSw2",
+                    roles: [
+                       {role: "readWrite", db: "dashboarddb"}
+                            ]
+                    })
                 Successfully added user: {
                   "user" : "dashboarduser",
                   "roles" : [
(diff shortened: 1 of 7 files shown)


Setup.md
```diff
@@ -79,6 +79,56 @@ You can pick and choose which collectors are applicable for your DevOps toolset
...
+
+Additionally, set the following export variables:
+
+```bash
+# Add these variables for authenticated proxy, required
+# for automated testing through an authenticated proxy:
+export HTTPAUTH_USER=companyId999
+export HTTPAUTH_PASS=yourPassword
+
+export HTTP_PROXY=http://your.proxy.domain.name:8080
```
jdamick: The HTTP/S_PROXY variables should look like:

```
HTTP_PROXY=http://<id>:<passwd>@proxy.somewhere:80
```

Setup.md
```diff
@@ -79,6 +79,56 @@ You can pick and choose which collectors are applicable for your DevOps toolset
...
+ </proxies>
+...
+```
+
+Additionally, set the following export variables:
+
+```bash
+# Add these variables for authenticated proxy, required
+# for automated testing through an authenticated proxy:
+export HTTPAUTH_USER=companyId999
```
jdamick: these are set in the HTTP/S_PROXY

core/src/test/java/com/capitalone/dashboard/repository/EmbeddedMongoDBRule.java
```diff
@@ -40,10 +42,28 @@
         public Proxy createProxy() {
 
             String proxy = System.getenv("HTTP_PROXY");
+            final String authUser = System.getenv("HTTPAUTH_USER");
```
jdamick: not necessary, should be in http_proxy

core/src/test/java/com/capitalone/dashboard/repository/EmbeddedMongoDBRule.java
```diff
@@ -40,10 +42,28 @@
...
+            final String authPassword = System.getenv("HTTPAUTH_PASS");
+            
             if (proxy == null || proxy.isEmpty()) {
                 proxy = System.getProperty("HTTP_PROXY");
             }
             try {
+            	// Case for Proxy authentication required
+                if ((proxy != null && !proxy.isEmpty()) && (authUser != null && !authUser.isEmpty()) && (authUser != null && !authPassword.isEmpty())) {
```
jdamick: can pull user & passwd from URL.getUserInfo()

amp5208:
I can make those modifications; however, please note, as Hygieia is currently written in master will not work with an authenticated proxy - this fix does resolve the issue, but I can make it pull out the username password from the default proxy variable strings.

Please do not close this PR, as I can commit an update directly to this PR.

amp5208:
I have pushed an update to this commit which utilizes the feedback that you provided, @jdamick .  Please review and let me know if you have any additional thoughts or concerns.  I also added a pom fix that was causing an issue to VersionOneFeatureCollector that @tabladrum / @amitmawkin saw the other day, and fixed a method declaration issue with the Core's Component's class (using a 'replace' instead of a 'put', which also performs a replace action).

amp5208:
@softwareklinic, can you confirm if the updated changes in this pull request will negatively/positively affect a *.pac proxy script, as you have noted as a concern in an earlier pull request?

core/src/test/java/com/capitalone/dashboard/repository/EmbeddedMongoDBRule.java
```diff
@@ -40,16 +44,38 @@
...
 
             String proxy = System.getenv("HTTP_PROXY");
+
             if (proxy == null || proxy.isEmpty()) {
                 proxy = System.getProperty("HTTP_PROXY");
             }
             try {
+            	URL proxyUrl = new URL(proxy);
+                StringTokenizer tokenizedUrl = new StringTokenizer(proxyUrl.getUserInfo().toString(),":");
+                final String authUser = tokenizedUrl.nextToken();
```
jdamick: this will throw for those of use that have a proxy with no user credentials..

core/src/test/java/com/capitalone/dashboard/repository/EmbeddedMongoDBRule.java
```diff
@@ -69,7 +95,7 @@ public void before() throws Throwable {
                     .net(new Net(port, Network.localhostIsIPv6())).build();
 
         Command command = Command.MongoD;
-        IRuntimeConfig runtimeConfig =
+		IRuntimeConfig runtimeConfig =
```
jdamick: formatting issue..

versionone-feature-collector/pom.xml
```diff
@@ -92,7 +89,6 @@
 			<groupId>commons-io</groupId>
 			<artifactId>commons-io</artifactId>
 			<version>2.4</version>
-			<scope>test</scope>
```
jdamick: why?

jdamick:
pac files aren't used, this only for env variables.

amp5208:
I'll fix the exception case, the formatting issues and resubmit a commit again.
the removal of scope "test" is actually fixing a runtime issue with VersionOneFeatureCollector, stopping it from actually running.  If you build on your machine with that change, you will see that you will now get the expected "$query" error that i am ultimately fixing.  It is required for that fix.

amp5208:
I have just committed the requested changes.  @jdamick , Please review and let me know if anything further should be done regarding the embedded mongodb proxy authentication enhancement

core/src/test/java/com/capitalone/dashboard/repository/EmbeddedMongoDBRule.java
```diff
@@ -24,89 +24,134 @@
...
+								return new PasswordAuthentication(authUser,
+										authPassword.toCharArray());
+							}
+						});
+
+						System.setProperty("http.proxyUser", authUser);
+						System.setProperty("http.proxyPassword", authPassword);
+
+					}
+				} catch (NullPointerException | IllegalArgumentException e) {
```
jdamick: dont catch NPEs, this is one of those code smells..  so revert back to way it was but just see if 	hasMoreTokens() is true before trying to nextToken(), that way it will support if there is a username and no pass cases too.

amp5208:
I see what you mean - i'm swallowing the exception instead of handling it before it ever happens.  I'll recommit next week with the changes you have suggested.  Thanks again for the input!

amp5208:
I have submitted some changes to the committed embedded MongoDB test case which check for authenticated proxy user info content, based on feedback from @jdamick .  I have left one catch for malformed exceptions, which does not entirely swallow the exception, as it prints a log of important information, but i have removed the swallowing NPE.  instead, i replaced this with a double-IF block that uses the tokenizer's built-in boolean check to see if more tokens are available before actually setting them.  this should cover for the use case that the username is blank (assumes then that password is also blank), and if the username is filled but the password is blank.  both cases will skip then to the proxy-only (non-auth) section.

Let me know if you have any remaining issues and i would be happy to fix them.

jdamick:
thanks

PR #256, opened 2015-12-28
Integration/unit tests for Stash/Bitbucket collectors/clients

srid99:
Added test cases for Stash and Bitbucket clients.

Do like to mention I found few things while writing the test case that there are few things should be fixed/improved. I added TODO's in the places where I found things has to be improved. I didn't fix them since I wanted to keep the implementation as it is and only getting the unit tests first so that we can refactor later (happily). Please go through the TODO's and let me know your opinion. Again I am fine with fixing them. Feel free to ping me in gitter if something is not clear or if I did something stupid (possibly)...

### DIFF
--- a/stash-scm-collector/pom.xml
+++ b/stash-scm-collector/pom.xml
@@ -1,106 +1,107 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
-	<modelVersion>4.0.0</modelVersion>
-	<groupId>com.capitalone.dashboard</groupId>
-	<artifactId>stash-scm-collector</artifactId>
-	<packaging>jar</packaging>
-	<name>${project.groupId}:${project.artifactId}</name>
-	<description>Stash Collector Microservice collecting stats from Stash</description>
-	<url>https://github.com/capitalone/Hygieia</url>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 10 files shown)


stash-scm-collector/pom.xml
```diff
@@ -1,106 +1,107 @@
...
-	</properties>
+        <testResources>
+            <testResource>
+                <directory>src/test/resources</directory>
+            </testResource>
+        </testResources>
+    </build>
 
+    <properties>
+        <java.version>1.7</java.version>
```
jdamick: we should be up on java8

stash-scm-collector/pom.xml
```diff
@@ -1,106 +1,107 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-mongodb</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
+            <version>4.1.0.RELEASE</version>
```
jdamick: dont lock to a specific ver, i think the base as this as property

stash-scm-collector/src/test/java/com/capitalone/dashboard/collector/DefaultStashClientTest.java
```diff
@@ -0,0 +1,192 @@
...
+    // TODO Throwing NPE doesn't seem right. Maybe this scenario (implementation) has to be handled properly.
+    @Test(expected = NullPointerException.class)
+    public void unexpectedResponse() {
+        server.response("{}");
+
+        stashClient.getCommits(gitRepo(), true);
+    }
+
+    // TODO Throwing NPE doesn't seem right. Maybe this scenario (implementation) has to be handled properly.
+    @Test(expected = NullPointerException.class)
```
jdamick: agreed, shouldnt be NPE

MarkRx:
Does this add the ability to choose which implementation is used?

PR #259, opened 2015-12-29
API Layer Health check on UI

amitmawkin:
Added a visual health check to depict if the UI can connect to api layer , it tries to connect to /api/dashboard  and check on returned HTTP status code, also modified the README.md for UI to reflect the same


jdamick:
this seems unnecessary to me, why not just make sure that if they try to do things that the page correctly notifies of an error reaching the api?

amitmawkin:
That is exactly what is this page doing, there is no other message that is shown today and this seems a nice addon telling developers that there is some problem reaching the api. There have been several issues opened by developers for the same  where api layer was down or unreachable

PR #278, opened 2016-01-16
Collect name of physical artifact that was deployed. 

tabladrum:
Component name is not good enough for tracking actual artifact. Other refactors.

### DIFF
--- a/udeploy-deployment-collector/src/main/java/com/capitalone/dashboard/collector/DefaultUDeployClient.java
+++ b/udeploy-deployment-collector/src/main/java/com/capitalone/dashboard/collector/DefaultUDeployClient.java
@@ -19,13 +19,17 @@ import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Component;
+import org.springframework.util.CollectionUtils;
 import org.springframework.web.client.RestClientException;
 import org.springframework.web.client.RestOperations;
 
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 @Component
 public class DefaultUDeployClient implements UDeployClient {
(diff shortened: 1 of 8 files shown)


udeploy-deployment-collector/src/main/java/com/capitalone/dashboard/collector/DefaultUDeployClient.java
```diff
@@ -144,55 +148,91 @@ public DefaultUDeployClient(UDeploySettings uDeploySettings,
  */
 
         // Failed to deploy list:
+        Set<String> failedComponents = getFailedComponents(nonCompliantResourceJSON);
+        Map<String, List<String>> versionFileMap = new HashMap<>();
+        for (Object item : allResourceJSON) {
```
jdamick: maybe if we load it into a model of some kind or a map it would be clearer whats going on rather than hand parsing it out.

udeploy-deployment-collector/src/main/java/com/capitalone/dashboard/collector/DefaultUDeployClient.java
```diff
@@ -144,55 +148,91 @@ public DefaultUDeployClient(UDeploySettings uDeploySettings,
...
-            JSONObject jsonObject = (JSONObject) item;
-            if (jsonObject == null) continue;
-            JSONArray childArray = getLowestLevelChildren(jsonObject, new JSONArray());
-            if (childArray.isEmpty()) continue;
-            for (Object child : childArray) {
-                JSONObject childObject = (JSONObject) child;
-                JSONArray jsonVersions = (JSONArray) childObject.get("versions");
-                if (jsonVersions == null || jsonVersions.size() == 0) continue;
+    private String cleanFileName(String fileName, String version) {
+        if (fileName.contains("-" + version))
```
jdamick: maybe regex instead would be easier..

udeploy-deployment-collector/src/main/java/com/capitalone/dashboard/collector/DefaultUDeployClient.java
tabladrum: Very hard as the json o/p from uDeploy is pretty generic and confusing.

udeploy-deployment-collector/src/main/java/com/capitalone/dashboard/collector/DefaultUDeployClient.java
tabladrum: Did a little test and RegEx in this case is lesser performing.

Issue #292, opened 2016-01-28
Problems using Jira with Proxy


oshingc:
When the fields for use of proxy are not fill the collector still using the null at the fields and errors are printed
![logs](https://cloud.githubusercontent.com/assets/3100819/12648240/ea656416-c5a6-11e5-9d8e-6afee1dd981a.png)

amp5208:
@oshingc , I see that you recently opened this issue, closed it, and reopened it:  Can I assume that you are still seeing this proxy issue with the Jira collector updates?  If so, I can submit a pull request to fix.

oshingc:
Yes please :D

amp5208:
Thank you for this!  Just made a pull request with a fix.  Please verify and let me know if you still see the issue form this pull:  https://github.com/capitalone/Hygieia/pull/293

oshingc:
Ok, you're welcome.

PR #293, opened 2016-01-28
Fixed based on proxy issue #292

amp5208:
As mentioned by user @oshingc, there is a bug in the proxy / auth proxy string validation with the Jira collector that needs to be resolved.  If a user has not set a proxy, but their property file has the proxy / auth proxy props set as blank, or not there at all, the tool will currently throw a URI syntax or null pointer exception, as indicated in the following issue:  https://github.com/capitalone/Hygieia/issues/292

This has now been fixed.  Please validate and merge as soon as possible.  I have also wrote additional test cases to cover these scenarios for Jira connections.

### DIFF
--- a/jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
+++ b/jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
@@ -3,7 +3,6 @@ package com.capitalone.dashboard.datafactory.jira;
 import java.io.IOException;
 import java.net.Authenticator;
 import java.net.InetSocketAddress;
-import java.net.MalformedURLException;
 import java.net.PasswordAuthentication;
 import java.net.Proxy;
 import java.net.URI;
@@ -28,6 +27,7 @@ import com.atlassian.util.concurrent.Promise;
 import com.google.common.collect.Lists;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
@@ -89,9 +89,8 @@ public class JiraDataFactoryImpl implements JiraDataFactory {
 	public JiraDataFactoryImpl(String jiraCredentials, String jiraBaseUrl, String jiraProxyUrl,
 			String jiraProxyPort) {
 		URI jiraUri;
-		if ((jiraBaseUrl != null) && (!jiraBaseUrl.isEmpty())) {
-			if ((jiraProxyUrl != null) && (jiraProxyPort != null) && (!jiraProxyUrl.isEmpty())
-					&& (!jiraProxyPort.isEmpty())) {
+		if (!StringUtils.isEmpty(jiraBaseUrl)) {
+			if (!StringUtils.isEmpty(jiraProxyUrl) && !StringUtils.isEmpty(jiraProxyPort)) {
 				jiraUri = this.createJiraConnection(jiraBaseUrl,
 						jiraProxyUrl + ":" + jiraProxyPort, this.decodeCredentials(jiraCredentials)
 								.get("username"),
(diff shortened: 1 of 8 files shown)


amp5208:
I am not using the URI constructors prior to my local constructor to support backwards compatibility.  If you look at my constructor, however, I am, in fact, directly leveraging the URI constructors.

StringUtils.isEmpty() did not work when I created this fix in this scenario for some reason.  However, let me try again and I will push an update.

jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
```diff
@@ -88,9 +89,21 @@ public JiraDataFactoryImpl(String jiraCredentials, String jiraBaseUrl) {
 	 */
 	public JiraDataFactoryImpl(String jiraCredentials, String jiraBaseUrl, String jiraProxyUrl,
 			String jiraProxyPort) {
-		URI jiraUri = this.createJiraConnection(jiraBaseUrl, jiraProxyUrl + ":" + jiraProxyPort,
-				this.decodeCredentials(jiraCredentials).get("username"),
-				this.decodeCredentials(jiraCredentials).get("password"));
+		URI jiraUri;
+		if ((jiraProxyUrl != null) && (jiraProxyPort != null) && (jiraProxyUrl != BLANK_LITERAL)
```
jdamick: just use Strings.isNullOrEmpty()

jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
```diff
@@ -145,9 +158,21 @@ public JiraDataFactoryImpl(int inPageSize, String jiraCredentials, String jiraBa
...
+		} else {
+			try {
+				LOGGER.debug("Handling without authenticated proxy (fields were available in properties settings but were blank)");
+				jiraUri = new URI(jiraBaseUrl);
+			} catch (URISyntaxException e) {
+				LOGGER.error("There was a problem reading the provide Jira base URI syntax");
+				jiraUri = null;
+			}
+		}
 		this.client = factory.createWithBasicHttpAuthentication(jiraUri,
```
jdamick: if jiraUri is null why do this?

amp5208:
I will switch to .isNullOrEmpty()

That section runs no matter what the user sets in their property file, and is run by the local clients.  For this reason, I have to validate all scenarios.  Did you have a suggestion on a better implementation here?

oshingc:
!= null definitely catch the case of a property with no value I tested that when proxyUrl was empty.
For now those validations are ok for solve the issue, 
Thanks @amp5208

amp5208:
Thank you @oshingc.  I have just pushed an update with the style suggestions from @tabladrum and @jdamick  above.  I also included additional connectivity checks for Jira.  Please validate once more for your configuration, if you are able.  Thank you!

oshingc:
It's working @amp5208

jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
```diff
@@ -145,12 +162,29 @@ public JiraDataFactoryImpl(int inPageSize, String jiraCredentials, String jiraBa
...
 	public JiraDataFactoryImpl(int inPageSize, String jiraCredentials, String jiraBaseUrl,
 			String jiraProxyUrl, String jiraProxyPort) {
-		URI jiraUri = this.createJiraConnection(jiraBaseUrl, jiraProxyUrl + ":" + jiraProxyPort,
-				this.decodeCredentials(jiraCredentials).get("username"),
-				this.decodeCredentials(jiraCredentials).get("password"));
-		this.client = factory.createWithBasicHttpAuthentication(jiraUri,
-				this.decodeCredentials(jiraCredentials).get("username"),
-				this.decodeCredentials(jiraCredentials).get("password"));
+		URI jiraUri;
+		if ((jiraBaseUrl != null) && (!jiraBaseUrl.isEmpty())) {
```
jdamick: can just Strings.isNullOrEmpty()  - we're pulling in guava, might as well use it..

jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
```diff
@@ -145,12 +162,29 @@ public JiraDataFactoryImpl(int inPageSize, String jiraCredentials, String jiraBa
...
+			} else {
+				try {
+					LOGGER.debug("Handling without authenticated proxy (fields were available in properties settings but were blank)");
+					jiraUri = new URI(jiraBaseUrl);
+				} catch (URISyntaxException e) {
+					LOGGER.error("There was a problem reading the provide Jira base URI syntax");
+					jiraUri = null;
+				}
+			}
+			this.client = factory.createWithBasicHttpAuthentication(jiraUri, this
```
jdamick: so if the uri is null why bother try, should probably just throw?

amp5208:
That's true I suppose, but PMD did not print any violations to fail the CI checks.  @jdamick , is it okay if I leave the extra parenthesis in my logic statements for the last push?

That's great to hear, and thank you for closing the ticket.  @jdamick and @tabladrum :  Once you are good with code style, you can merge this pull.

amp5208:
I was avoiding throwing null pointers and rather handle... but let me see what I can do.

amp5208:
Updated based on suggestions by @tabladrum and @jdamick .  Pending review.  Tests pass and all cases set before this commit should work the same.

amp5208:
I have also pushed an important fix related to jira connection timeouts under production load.  Please accept this pull request to also get this fix.

Issue #330, opened 2016-02-14
Error after pull #329 - bower install fails


jimzucker:
I tried building after the #329 was merged and I get an error, I am trying to research it..

[INFO] --- frontend-maven-plugin:0.0.27:bower (bower install) @ UI ---
[INFO] Found proxies: []
[INFO] Running 'bower install' in /Users/jaz/work/Hygieia/UI
[INFO] bower bootstrap-sweetalert#~0.4.5           cached git://github.com/lipis/bootstrap-sweetalert.git#0.4.5
[INFO] bower bootstrap-sweetalert#~0.4.5         validate 0.4.5 against git://github.com/lipis/bootstrap-sweetalert.git#~0.4.5
[INFO] bower regression-js#~1.3.0                  cached git://github.com/Tom-Alexander/regression-js.git#1.3.0
[INFO] bower regression-js#~1.3.0                validate 1.3.0 against git://github.com/Tom-Alexander/regression-js.git#~1.3.0
[INFO] bower ng-sortable#~1.3.2                    cached git://github.com/a5hik/ng-sortable.git#1.3.2
[INFO] bower ng-sortable#~1.3.2                  validate 1.3.2 against git://github.com/a5hik/ng-sortable.git#~1.3.2
[INFO] bower dexie#~1.2.0                          cached git://github.com/dfahlander/dexie.js.git#1.2.0
[INFO] bower dexie#~1.2.0                        validate 1.2.0 against git://github.com/dfahlander/dexie.js.git#~1.2.0
[INFO] bower operative#~0.4.4                      cached git://github.com/padolsey/operative.git#0.4.4
[INFO] bower operative#~0.4.4                    validate 0.4.4 against git://github.com/padolsey/operative.git#~0.4.4
[INFO] bower angular-chartist.js#~3.3.13           cached git://github.com/paradox41/angular-chartist.js.git#3.3.13
[INFO] bower angular-chartist.js#~3.3.13         validate 3.3.13 against git://github.com/paradox41/angular-chartist.js.git#~3.3.13
[INFO] bower chartist#~0.4.3                       cached git://github.com/gionkunz/chartist-js.git#0.4.4
[INFO] bower chartist#~0.4.3                     validate 0.4.4 against git://github.com/gionkunz/chartist-js.git#~0.4.3
[INFO] bower chartist#~0.9.5                       cached git://github.com/gionkunz/chartist-js.git#0.9.5
[...]

PR #329, opened 2016-02-13
Hygieia 2 product health pipeline and related functionality

jayhogan:
This pull request is the culmination of a 4 week development effort to build a product dashboard to display pipeline health across multiple teams. While this is the major feature delivered by this PR, there are also numerous enhancements and bug fixes, including:

## Enhancements

- Ability to create a product dashboard
- Enhance list view to improve filtering and make product dashboards visually distinct
- Ability to add/remove a team to a product dashboard
- Ability reorder the team rows in a product dashboard via drag and drop
- Ability to view pipeline health (i.e Quality) summary metrics for each team on the product dashboard
- Ability to view commit details for a given pipeline stage
- Ability to view quality details for a given team row
- New API to allow pushing of commits to Hygieia via GitHub webhooks
- New API to allow pushing of build data
- New API to allow pushing of binary artifacts
- New API to allow pushing of deploy data
- New API to allow pushing of quality data
- New API to allow pushing of test result data
- Introduction of a client-side database (Dexie) to reduce the amount data transferred from server to client on every request

## Bug fixes

- Fixes #265 by simplifying and improving the UI gulp build process
- Fixes #113 by creating a global config object in the UI to allow command line args to bypass defaults
- Fixes #275 by upgrading to the latest version of the Chartist library and fixing the existing charts accordingly
- Fixes #170 and #274 via creation of new Chartist plugin

hygieia2

Issue #323, opened 2016-02-09
Upgrade to Java 1.8


jimzucker:
We should upgrade the POMs from 1.7 to 1.8.  I will try the upgrade, if I dont get any objection/see any issues I will submit the pull request.

PR #332, opened 2016-02-16
Reference #323 - Upgraded to java 1.8 

jimzucker:
Clean branch for java 1.8 upgrade.

Changes:
Reference #323 - Upgraded to java 1.8 
Reference #330 - Error after pull #329 - bower install fails. Solved by adding explicit reference for chartist.


tabladrum:
@amp5208 - do you have any concerns about JIRA and VersionOne collectors not working with 1.8? I know we had some work to do to move to 1.7.

amp5208:
@tabladrum , I have already confirmed from my end that the feature collectors and APIs are compatible now with 1.8.

Issue #368, opened 2016-02-25
Code collector widget showing different tech debt value


papithag:
Hello
I am using SonarQube for code quality and coverage, technical debt value is varying in Sonarqube dashboard and in Hygieia code collector widget.

Is there any fix around so that I can use in Hygieia dashboard code?

jimzucker:
I ran a test using sonar 5.3 and the Hygieia code base and reproduced the issue.  It will need to be investigated.

Here is my review:
***Tech Debt: 6d on Hygieia and 15d in Sonar
% Coverage  - LGTM
#issues - LGTM


<img width="710" alt="screenshot 2016-02-25 16 47 29" src="https://cloud.githubusercontent.com/assets/14372687/13339701/7e72de7e-dbdf-11e5-95c1-cf0834ecffa1.png">

<img width="354" alt="screenshot 2016-02-25 16 41 55" src="https://cloud.githubusercontent.com/assets/14372687/13339647/172e195e-dbdf-11e5-8fe5-4dd372d0362c.png">
<img width="310" alt="screenshot 2016-02-25 16 41 47" src="https://cloud.githubusercontent.com/assets/14372687/13339649/1923f094-dbdf-11e5-9078-4c4380621d70.png">

<img width="1160" alt="screenshot 2016-02-25 16 41 20" src="https://cloud.githubusercontent.com/assets/14372687/13339650/192aeb2e-dbdf-11e5-8892-4525324ba7b8.png">

papithag:
Hi @jimzucker , 
any update to this issue?

jimzucker:
@papithag I took a look at the issue it the calc in the dashboard

There is a method, calculateTechnicalDebt.  I don think the calculation for Day and year are correct, the is what I see 

A day is counted as 1440 min, I think it should be 480.

Year is counted as 525600 minutes I think it should be 105,600 minutes. This is what I think it should be:
 Min 	          Hours 	     Days 	 Year 
 60.000 	         1.000 	     0.125 	 0.001 
 480.000 	          8.000 	     1.000 	 0.005   *assume a day == 8 hours
 105,600.000 	 1,760.000     220	 1.000  * assume a year is 220 workdays

I am testing the patch now...

papithag:
Hi @jimzucker 
Awaiting for the fix :-)

jimzucker:
To get the #s to tie to sonar I had to use 8.25 hours/day.

 Min 	          Hours 	     Days 	 Year 
[...]

PR #375, opened 2016-02-29
Tech debt incorrect

jimzucker:
reference #368 Code collector widget showing different tech debt value

In the current code I found 2 issues:

#1 the values in the DB returned in the rest call is in minutes, so we should be diving by 8*60 for 1d. (maybe older versions of sonar are using seconds if we find that is the case we may have to make this configurable)

#2 it appears the sonar ui is using floor() vs ceil() of the calculation, so I changed that to line up the UI's.

### DIFF
--- a/UI/src/components/widgets/codeanalysis/view.js
+++ b/UI/src/components/widgets/codeanalysis/view.js
@@ -52,7 +52,9 @@
             ctrl.rulesCompliance = getMetric(caData.metrics, 'violations_density');
 
             ctrl.technicalDebt = getMetric(caData.metrics, 'sqale_index');
-            ctrl.technicalDebt.formattedValue = calculateTechnicalDebt(ctrl.technicalDebt.value);
+	 
+            //ctrl.technicalDebt.formattedValue = calculateTechnicalDebt(ctrl.technicalDebt.value);
+            ctrl.technicalDebt.formattedValue = ctrl.technicalDebt.formattedValue;
 
             ctrl.linesofCode = getMetric(caData.metrics, 'ncloc');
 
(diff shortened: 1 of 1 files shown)


tabladrum:
We have been wondering abt it for a while. Thanks for digging deep. Older version of Sonar was okay.

jimzucker:
changed to display formatted value returned by sonar instead of computing from minutes, this will ensure www are as consistent as practical with the Sonar UI.

UI/src/components/widgets/codeanalysis/view.js
```diff
@@ -52,7 +52,9 @@
             ctrl.rulesCompliance = getMetric(caData.metrics, 'violations_density');
 
             ctrl.technicalDebt = getMetric(caData.metrics, 'sqale_index');
-            ctrl.technicalDebt.formattedValue = calculateTechnicalDebt(ctrl.technicalDebt.value);
+	 
+            //ctrl.technicalDebt.formattedValue = calculateTechnicalDebt(ctrl.technicalDebt.value);
+            ctrl.technicalDebt.formattedValue = ctrl.technicalDebt.formattedValue;
```
jdamick: this is not doing what the comment states

jdamick:
@jimzucker can you verify and recreate, this doesnt seem correct

jimzucker:
@jdamick - I see the comment on the pull request was outdated, but the comment not he issue #368 & the commit are OK.  I updated the pull request description, if you want a new pull request let me know.

jdamick:
but this part:
```ctrl.technicalDebt.formattedValue = ctrl.technicalDebt.formattedValue;```
why?

jimzucker:
@jdamick - yes something looks wrong, let me check my branches, thanks for catching this.

jimzucker:
@jdamick - OK I see what happened, I just didn't pay attention to the left side of the equation when I made the fix, I commented it out as it is a NOP, when the JSON is parsed the value is set correctly. I will submit a new pull request.

Issue #352, opened 2016-02-21
PMD Warnings - Avoid catching NullPointerException; consider removing the cause of the NPE (14 cases)


jimzucker:
This needs input from the team, do we want to fix these or configure PMD to just ignore them?

Description Resource    Path    Location    Type
Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 77 PMD Marker

Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 119    PMD Marker

Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 172    PMD Marker

Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 180    PMD Marker

Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 190    PMD Marker

Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 202    PMD Marker

Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 255    PMD Marker

Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 308    PMD Marker

Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 361    PMD Marker

Avoid catching NullPointerException; consider removing the cause of the NPE.    FeatureServiceImpl.java /api/src/main/java/com/capitalone/dashboard/service line 415    PMD Marker

jimzucker:
@tabladrum @jdamick  - What do we want to do with this category?  Maybe we should just disable it the PMD rules and allow it, seems like make-work to keep surpassing them, do you really want to re-write such code?

tabladrum:
I really want @amp5208 fix this code.

amp5208:
[...]

Issue #343, opened 2016-02-21
Type safety Java Error in eclipse (10 cases)


jimzucker:
We should clean this up so a developer can see new issues when changing code.

Description	Resource	Path	Location
Type safety: The method add(Object) belongs to the raw type ArrayList. References to generic type ArrayList<E> should be parameterized	DefaultUDeployClient.java	/udeploy-deployment-collector/src/main/java/com/capitalone/dashboard/collector	line 242
Type safety: The expression of type Map needs unchecked conversion to conform to Map<String,String>	Dashboard.java	/core/src/main/java/com/capitalone/dashboard/model	line 84
Type safety: The expression of type Map needs unchecked conversion to conform to Map<? extends String,? extends String>	PipelineServiceImpl.java	/api/src/main/java/com/capitalone/dashboard/service	line 106
Type safety: The expression of type List needs unchecked conversion to conform to List<PipelineResponse>	PipelineServiceTest.java	/api/src/test/java/com/capitalone/dashboard/service	line 70
Type safety: The expression of type Collection needs unchecked conversion to conform to Collection<ObjectId>	CollectorServiceTest.java	/api/src/test/java/com/capitalone/dashboard/service	line 36
Type safety: The constructor HttpEntity(MultiValueMap) belongs to the raw type HttpEntity. References to generic type HttpEntity<T> should be parameterized	DefaultHudsonClientTests.java	/jenkins-build-collector/src/test/java/com/capitalone/dashboard/collector	line 105
Type safety: The constructor HttpEntity(MultiValueMap) belongs to the raw type HttpEntity. References to generic type HttpEntity<T> should be parameterized	DefaultHudsonClientTests.java	/jenkins-build-collector/src/test/java/com/capitalone/dashboard/collector	line 119
Type safety: The constructor HttpEntity(MultiValueMap) belongs to the raw type HttpEntity. References to generic type HttpEntity<T> should be parameterized	DefaultHudsonClientTests.java	/jenkins-build-collector/src/test/java/com/capitalone/dashboard/collector	line 133
Type safety: The constructor HttpEntity(MultiValueMap) belongs to the raw type HttpEntity. References to generic type HttpEntity<T> should be parameterized	DefaultJenkinsClientTest.java	/jenkins-cucumber-test-collector/src/test/java/com/capitalone/dashboard/collector	line 74
[...]

Issue #350, opened 2016-02-21
Eclipse warnings cleanup - Resource leak (1 Case)


jimzucker:
This should be reviewed and addressed.

Resource leak: '<unassigned Closeable value>' is never closed	MarkdownTest.java	/core/src/test/java/com/capitalone/dashboard	line 63

jimzucker:
@jdamick  - can you take a look at this one, it was introduced with your commit in Dec.

tabladrum:
Merged.

PR #396, opened 2016-03-04
Eclipse  Type safety Java Error cleanup

jimzucker:
reference #343 - Type safety Java Error in eclipse (10 cases)
    There are a few cases left that need more research, the ones in the JIRA collectors broke tests when I fixed them.  The other 2 cases need more research on the proper resolution.

reference #352 - PMD Warnings - Avoid catching NullPointerException; consider removing the cause of the NPE - For now suppresed and added to TODO list
            @tabladrum noted that these should be fixed, for today I suppressed them and marked them for the TODO list to cleanup, in this way we can monitor PMD more effectively.

screenshot 2016-03-04 11 38 41

### DIFF
--- a/core/src/test/java/com/capitalone/dashboard/MarkdownTest.java
+++ b/core/src/test/java/com/capitalone/dashboard/MarkdownTest.java
@@ -11,6 +11,7 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.charset.Charset;
 import java.util.Locale;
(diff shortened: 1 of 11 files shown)


jimzucker:
Looks like I missed a checkin, I will fix it.. give me a few minutes.

core/src/test/java/com/capitalone/dashboard/MarkdownTest.java
```diff
@@ -60,9 +61,18 @@ private File getProjectRoot() {
...
+    	FileInputStream input = null;
+        try {
+        	input = new FileInputStream(filename);
+        	FileChannel channel = input.getChannel();
             ByteBuffer buffer = channel.map(FileChannel.MapMode.READ_ONLY, 0, (int) channel.size());
-            return Charset.forName("8859_1").newDecoder().decode(buffer);
+            CharBuffer result = Charset.forName("8859_1").newDecoder().decode(buffer);
+            input.close();
+			return result;
+        } catch (Exception e) {
```
jdamick: i dont understand why you changed this?

jimzucker: See #350  - Eclipse identified we where not close the FileChannel resource returned, so my goal was to fix that.

This was the original code:

    CharSequence fromFile(String filename) throws IOException {
        try (FileChannel channel = new FileInputStream(filename).getChannel()) {
            ByteBuffer buffer = channel.map(FileChannel.MapMode.READ_ONLY, 0, (int) channel.size());
            return Charset.forName("8859_1").newDecoder().decode(buffer);
        }
    }

jimzucker:
@jdamick are you OK to accept this pull request now?

PR #409, opened 2016-03-09
reference #368 Code collector widget showing different tech debt value vs SONAR UI

jimzucker:
@jdamick I removed unnecessary code identified in code review

reference #368 Code collector widget showing different tech debt value

In the current code I found the math was wrong, but after further testing we found it better to simply use the formatted value returned in the JSON from SONAR, then we will always match.


jdamick:
ok makes sense, thanks

jimzucker:
@jdamick I think this conflicts due to the merge/revert, do you want me to create a clean branch and resolve the merge or will you take care of it?

jimzucker:
created a new branch to resolve conflict -  #435

PR #419, opened 2016-03-11
Fixed VersionOne Authentication Error

amp5208:
Currently with the VersionOne feature collector, if you are successfully able to communicate with VersionOne, but your access token is not authorized, VersionOne does not send a standard HTTP 403 response; instead, it responds with a custom message body <em>{"error":"Unauthorized"}</em>.  Currently, the VersionOneFeatureCollector will interpret this as a ClassCastException, because the response is not in the standard JSONArray format normally provided by a successful VersionOne API response.

This fix will determine if the response was validly responding with corrupted data, or if it is a custom VersionOne Not Authorized response.  This will log at error level accordingly at runtime.


amp5208:
Yes, but the exception is not verbose, and does not accurately inform the user that it was their authentication that failed.  You can replicate this scenario by taking the current master of Hygieia's VersionOneFeatureCollector, then use a bad VersionOne authentication token... it will give you a nondescript "ClassCastException" error, resulting in a JSONArray casting error message.  It is incorrect - my fix shows instead that you received a VersionOne unauthorized exception.

I can remove the global Exception catch here.  Is that acceptable?  Remember:  This quick fix does not include the PMD fixes planned for next iteration (which would include fixes arbitration NullPointerException catches, etc.).

amp5208:
Closing, as I will be submitting a new request pushing to @tabladrum 's v1c-refactor branch

PR #464, opened 2016-03-23
Fix for orphaned commits in commit stage of pipeline when no build configured

jayhogan:
This PR is to fix a bug whereby commits can be orphaned in the commit stage of the product pipeline widget.

Here is the scenario:

1. A (team) dashboard exists with a configured SCM widget (using the Github webhook push method)
2. The same dashboard does **NOT** have a build widget configured
3. A new commit (call it commit A) is pushed to the Github repo
4. The build widget is then added to the team dashboard
5. A new product dashboard is created that references the team dashboard.
6. Commit A is displayed in the commit stage and never moves to subsequent stages.

This PR proposes to fix the bug by not adding commits to the pipeline of a team dashboard if there is no build widget configured.

### DIFF
--- a/core/src/main/java/com/capitalone/dashboard/event/CommitEventListener.java
+++ b/core/src/main/java/com/capitalone/dashboard/event/CommitEventListener.java
@@ -28,13 +28,19 @@ public class CommitEventListener extends HygieiaMongoEventListener<Commit> {
     public void onAfterSave(AfterSaveEvent<Commit> event) {
         Commit commit = event.getSource();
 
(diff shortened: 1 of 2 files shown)


tabladrum:
Should we consider preventing the creation of a product dashboard if the following widgets are not configured on a team dashboard:
Code Repo, Build, Deploy and Pipeline?

jayhogan:
I thought of that and do consider it an option. I wonder though if all team dashboards will have a Deploy widget. It seems to me that a team could contribute code, perhaps via a branch and never actually deploys to an external environment (beyond local development). That begs the question if that development should be tracked on a Product dashboard?

PR #506, opened 2016-04-20
Changed the model classes - Added lombok.

sheosinha:
Hey, Added lombok annotations to the model classes and removed the getter and setter methods.


sheosinha:
@tabladrum and @amitmawkin - I removed the getters and setters from all the model classes. The pmd thinks that the private variables are not used at all because the public methods for accessing those variables are wrapped behind the lombok. This will need the pmd xml change.

PR #530, opened 2016-04-28
Cloud Widget changes

amitmawkin:
Fixed bug in cloud instance data refresh
Fixed Subnet Datatypes


core/src/main/java/com/capitalone/dashboard/request/CloudSubnetCreateRequest.java
```diff
@@ -23,7 +23,7 @@
     private long creationDate;
     private long lastUpdateDate;
     private List<NameValue> tags = new ArrayList<>();
-    private List<NameValueCount> ipUsage = new ArrayList<>();
+    private Map<String, Integer> ipUsage;
```
tabladrum: what will be value in the String in the map?

sheosinha:
String will be the name of the resource where that particular subnet is being attached and the value will be the number of Ips consumed by that resource/s. It is coming from that tag name for Instances.

PR #492, opened 2016-04-10
Added Hygieia.pptx

softwareklinic:
This file outlines Hygieia components/collectors/properties files etc... hope it will be useful to newbies

PR #537, opened 2016-05-02
add the overview in Markdown instead of PPTX 

azat-co:
add the overview in Markdown instead of PPTX CC @jdamick @softwareklinic #492


azat-co:
😄

azat-co:
@tabladrum thanks!

PR #581, opened 2016-05-19
Added better error message for 401 errors from JIRA

jimzucker:
Added better error message for 401 error, handled /n added by mac base64 to make more resiliant and updated jenkins test server docker file

### DIFF
--- a/jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
+++ b/jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
@@ -2,6 +2,7 @@ package com.capitalone.dashboard.datafactory.jira;
 
 import com.atlassian.jira.rest.client.api.JiraRestClient;
 import com.atlassian.jira.rest.client.api.JiraRestClientFactory;
+import com.atlassian.jira.rest.client.api.RestClientException;
 import com.atlassian.jira.rest.client.api.domain.BasicProject;
 import com.atlassian.jira.rest.client.api.domain.Issue;
 import com.atlassian.jira.rest.client.api.domain.SearchResult;
@@ -254,8 +255,20 @@ public class JiraDataFactoryImpl implements JiraDataFactory {
 				}
 			} catch (Exception e) {
 				issues = new ArrayList<BasicProject>();
-				LOGGER.warn("No result was available from Jira unexpectedly - defaulting to blank response. The reason for this fault is the following:"
-						+ e.getCause());
+				if (e.getClass() == RestClientException.class) {
+					RestClientException cause = (RestClientException) e.getCause();
+					if (cause.getStatusCode().get() == 401 ) {
+						LOGGER.error("Error 401 connecting to JIRA server, your credentials are probably wrong. Note: Ensure you are using JIRA user name not your email address.");
+					} else {
+						LOGGER.error("No result was available from Jira unexpectedly - defaulting to blank response. The reason for this fault is the following:"
+								+ e.getCause());
+					}
+					throw e;
+
+				} else {
+					LOGGER.error("No result was available from Jira unexpectedly - defaulting to blank response. The reason for this fault is the following:"
+							+ e.getCause());
+				}
 			}
 		} else {
 			issues = null;
@@ -325,8 +338,9 @@ public class JiraDataFactoryImpl implements JiraDataFactory {
 	protected Map<String, String> decodeCredentials(String jiraBasicAuthCredentialsInBase64) {
 		Map<String, String> credMap = new LinkedHashMap<String, String>();
 		if (jiraBasicAuthCredentialsInBase64 != null) {
+				//the tokenize includes a \n to ensure we trim those off the end (mac base64 adds these!)
 			StringTokenizer tokenizer = new StringTokenizer(new String(
-					Base64.decodeBase64(jiraBasicAuthCredentialsInBase64)), ":");
+					Base64.decodeBase64(jiraBasicAuthCredentialsInBase64)), ":\n");
 			for (int i = 0; tokenizer.hasMoreTokens(); i++) {
 				if (i == 0) {
 					credMap.put("username", tokenizer.nextToken());
@@ -335,7 +349,7 @@ public class JiraDataFactoryImpl implements JiraDataFactory {
 				}
 			}
 		}
-
+		
 		return credMap;
 
 	}
--- a/test-servers/jenkins/Dockerfile
+++ b/test-servers/jenkins/Dockerfile
@@ -1,4 +1,4 @@
-FROM jenkins:1.642.1
+FROM jenkins:latest
 
 #based on MAINTAINER Maxfield Stewart - github: maxfields2000/dockerjenkins_tutorial
 


jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
```diff
@@ -325,8 +338,9 @@ public int getPageSize() {
 	protected Map<String, String> decodeCredentials(String jiraBasicAuthCredentialsInBase64) {
 		Map<String, String> credMap = new LinkedHashMap<String, String>();
 		if (jiraBasicAuthCredentialsInBase64 != null) {
+				//the tokenize includes a \n to ensure we trim those off the end (mac base64 adds these!)
 			StringTokenizer tokenizer = new StringTokenizer(new String(
-					Base64.decodeBase64(jiraBasicAuthCredentialsInBase64)), ":");
+					Base64.decodeBase64(jiraBasicAuthCredentialsInBase64)), ":\n");
```
amp5208: Will this not add an additional string token by doing this (e.g., a username, password, and a potential other token with a return carriage in it)?  The next section of code (lines 330-336) will loop through the tokens; please ensure this scenario will not cause a downstream ArrayIndexOutOfBoundsException; at minimum, please prove this with a basic JUnit test (note: there is already a small test that evaluates this class which you can append to).

test-servers/jenkins/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM jenkins:1.642.1
+FROM jenkins:latest
```
amp5208: @amitmawkin | @jdamick  | @tabladrum :  Do we want to allow the Dockerfile to use latest Jenkins image, or does Hygieia require the specified version of the Jenkins image?

jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
```diff
@@ -254,8 +255,20 @@ public String setQuery(String query) {
 				}
 			} catch (Exception e) {
 				issues = new ArrayList<BasicProject>();
-				LOGGER.warn("No result was available from Jira unexpectedly - defaulting to blank response. The reason for this fault is the following:"
-						+ e.getCause());
+				if (e.getClass() == RestClientException.class) {
```
amp5208: This would definitely work; however, I believe that Hygieia now comes with its own Exception class that we can implement here.

That being said, if there is currently no other implementation of the Hygieia Exception class within this class, don't bother editing it; I will update the classes properly with my next PR.

amp5208:
@jimzucker , Thank you for your proposed fix.  I have included some in-line comments; please address the TODOs and check the check boxes when verified/changed.

@tabladrum / @jdamick :  I have confirmed that @jimzucker's fix is required, as it resolves some confusion caused by the Mac implementation of base64 encoding, commonly used to generate the basic authentication username:password combo for the JiraFeatureCollector.  Once the above-noted check boxes are addressed, I am okay with this pull request.  This will also not derail my currently feature development.

test-servers/jenkins/Dockerfile
jimzucker: Since this is intended only as a test utility I think latest is OK to make it convent and reduce maintenance, I found the version I was using has disappeared.  There is no risk to the project, worse that will happen is the utility won’t work..

I can pin it and maybe build a unit test but I dont think it is worth the effort..

thoughts?


On May 19, 2016, at 2:29 PM, Arnold Pelka <notifications@github.com<mailto:notifications@github.com>> wrote:


In test-servers/jenkins/Dockerfile<https://github.com/capitalone/Hygieia/pull/581#discussion_r63954615>:

> @@ -1,4 +1,4 @@
> -FROM jenkins:1.642.1
> +FROM jenkins:latest


@amitmawkin<https://github.com/amitmawkin> | @jdamick<https://github.com/jdamick> | @tabladrum<https://github.com/tabladrum> : Do we want to allow the Dockerfile to use latest Jenkins image, or does Hygieia require the specified version of the Jenkins image?

amp5208: Your statement makes sense to me; @tabladrum | @jdamick :  Any thoughts?

tabladrum:
@jimzucker @amp5208 can we do this PR to master? It is requesting merge to a feature branch.

amp5208:
Agreed - I missed that.  @jimzucker , please modify PR to push to Hygieia upstream master.  Also, be sure that your changes work with the modifications going into PR #606

jira-feature-collector/src/main/java/com/capitalone/dashboard/datafactory/jira/JiraDataFactoryImpl.java
```diff
@@ -254,8 +255,20 @@ public String setQuery(String query) {
 				}
 			} catch (Exception e) {
```
amp5208: @amitmawkin , Yes, seriously.  Unfortunately, when claiming the Jira promise in this method, there does not seem to be a way to actually get even the Jira-level exception - please feel free to try if you do not believe me.  The Jira exception propagated here does not seem to be accessible.

You can validate when this exception is thrown by ensuring that a message is never received by the Jira collector from the source system.  However, this is not an artifact of @jimzucker's PR:  Please move this to a new issue in GitHub if you request that it be modified.

amitmawkin: What kind of exception are you seeing ?

. Did we open an issue with Jira support...?

amp5208: @amitmawkin :  I just re-checked my statement above to get the odd exception behavior - upon retrospective, I am, in fact, able to catch the exception I thought previously uncatchable:

```java
com.atlassian.jira.rest.client.api.RestClientException
```

I will add this fix to my next commit on this PR and fix.  If we do see an issue going forward where an uncatchable exception is being thrown, i will follow up with Jira support - I hadn't thought of that originally.

amp5208: @amitmawkin , **Correction**:  I have made the exception update to my PR.  Please review my latest commit change:  https://github.com/capitalone/Hygieia/pull/606/commits/0eb9af2a63f0c580d5d2fa793e7a5e15d54d3ee1 .  Feel free to move the discussion over to that PR.

amp5208: @amitmawkin , I take that back again... I was right - Jira's RestClientException does not also throw or handle the following exception:  ```java.net.UnknownHostException```.  It is thus also uncatchable - I am nearly forced to throw it on my method implementation, myself.  You can see this in the failed build Travis.ci build for the above-noted commit.

However, I was able to handle it myself in a different way - I am now directly validating if a Jira host provided is an available namespace before ever attempting to resolve any query promises.  Please review my code at the following commit/PR:  https://github.com/capitalone/Hygieia/pull/606/commits/60af6cb2ccdf926a8318a1e26938b7b4e7b94114

amitmawkin:
LGTM

Issue #513, opened 2016-04-22
Please add more attributes/information to the deploy widget


rashisaxena:
At the moment, deploy widget gives us the information about Environment and by clicking on it it takes us to the artefact details such as artefact name, version and status of the build.

However we need more params  especially the release baseline number, deployment timelines and link to release notes; without that the information the deploy gadget is incomplete to the management and to the release management teams.

tabladrum:
We now have it as a backlog. Will work on it during last week of May.

rashisaxena:
Hi,

sure, I am waiting for this to be implemented. :)

Regards,
Rashi

MarkRx:
Does work on this include being able to select multiple applications at once in the deploy widget?

tabladrum:
No it does not.

PR #594, opened 2016-05-26
Creation of XebiaLabs XLDeploy collector

MarkRx:
Added XLDeploy collector (https://xebialabs.com/products/xl-deploy/). The project was based off of the UDeploy collector.

Note the following differences I found between XLDeploy and UDeploy:

- I don't gather "component" information since as far as I know in XLD components (called deployables) are deployed with an application package (version of an application) in an all-or-nothing manner. Instead on the UI each application will have a single component with the name of the application.
- I don't think XLD has a way to get environment health information (how many servers up/down) so instead that data isn't set. Each application will thus be listed as having 0 servers up and 0 servers down.
- Complex scenarios like transitive application dependencies and composite packages are not handled. I may add them in the future depending on project needs.

The collector works by parsing through application deployment history. New applications or applications that have not been enabled in the past day or so (since beginning of yesterday) parse through 3 months of history. For efficiency applications that have recently been parsed (since the beginning of yesterday) only hit the past day or so of history (since the beginning of yesterday). XLD acts strange in that it doesn't archive a task when it completes but instead when it is closed in the UI. For that reason I look back an extra day to pick up any history items that could of been missed.

I have signed the license agreement.


AdonisCelestine:
@MarkRx
Awesome...This saves us a lot of effort... We are using XLDeploy & XLRelease as well. Currently we are building a collector for XLRelease, wondering if you are also using XLRelease.

tabladrum:
@MarkRx Will merge this very soon. Thanks a lot for the PR.

MarkRx:
I have not thought about how XLR would fit in yet. I am currently looking at being able to track multiple XLD deployment packages at once by having the UI + api be able to select more than one app at once.

tabladrum:
Please see https://github.com/capitalone/Hygieia/issues/513 - 
Enhancement request to pull in some release related information.

PR #614, opened 2016-06-03
Allow use of custom spring properties configurations

MarkRx:
Change provides the ability to specify the spring config location when using a docker mount. Doing so allows for the setting of properties not defined as environment variables without having to build a new docker image. Using this mechanism is more secure as well since it prevents sensitive information leaking through docker inspect or a ps attack.

Example:

> docker run -t -p 8080:8080 -v /logs:/hygieia/logs -v "/vagrant/config:/hygieia/config" --name hygieia-api -e "SPRING_CONFIG_LOCATION=/hygieia/config/allconfig.properties" -i hygieia-api:latest

Note that missing the MONGO_PORT environment variable no longer terminates the container and the echo of MONGODB_HOST and MONGODB_PORT was removed. Since it wasn't consistent, has a default value, and is echo'd anyway it should be fine.

The SPRING_CONFIG_LOCATION defaults to the file that is generated by the property builder of the image if not specified.


MarkRx:
Update: learned about docker-compose in a comment from @jimzucker. Seems that it can override commands and environment variables. From what I've read a 'docker inspect' environment variable leak isn't considered serious because you need root to be able to run it and hence can already do what you want. Environment variables are not the most secure since they are exposed during linking so should only be used in a confined environment.

https://github.com/docker/docker/issues/13490
https://github.com/docker/docker/issues/5169

Not sure if this PR is useful or not. Maybe for people who don't run docker-compose? I'll wait and see if someone who is more familiar with docker has any opinions.

MarkRx:
Closing as this is unnecessary - commands can be overridden with docker compose to do whatever is needed. Example:
hygieia-jira-feature-collector:
  image: hygieia-jira-feature-collector:latest
  container_name: hygieia-jira
  volumes:
  - ./logs:/hygieia/logs
  - /vagrant/config:/hygieia/config
  links:
  - hygieia-api
  command: ["/bin/sh", "-c", "egrep -vi 'password|Credentials' '/hygieia/config/jira-docker.properties' && java -jar /hygieia/jira-feature-collector*.jar --spring.config.location=/hygieia/config/jira-docker.properties"]
  environment:
  - MONGO_PORT=tcp://192.168.33.11:27017

PR #629, opened 2016-06-17
Invalid json causing errors while testing

RobertKelly:
The changes to this file made on April 19 2016 cause JS errors in UI testing. You are unable to use the dashboard UI due to bower errors parsing invalid json. The bracket at the beginning and the comma + brace at the end need to be removed so that the UI runs while using gulp serve to test changes.


RobertKelly:
Don't see how my change caused that build error.  Looks like there's a missing file error and I didn't remove that from the workspace.  Working fine here.

RobertKelly:
Reopening after changes in master to fix build

Issue #635, opened 2016-06-21
loadash upgrade casues some functions not working


tabladrum:
bower.js:13113 TypeError: _.findWhere is not a function
    at createCollectorItem (config.js:108)
    at RepoConfigController.submitForm [as submit] (config.js:88)
    at $parseFunctionCall (bower.js:13881)
    at callback (bower.js:23107)
    at Scope.$eval (bower.js:15977)
    at Scope.$apply (bower.js:16076)

See https://github.com/angular-ui/angular-google-maps/issues/1682

rodrigosiviero:
Just tried the master and it is working, thanks again!

PR #637, opened 2016-06-21
lodash version back to 2.4.1

tabladrum:


tabladrum:
See https://github.com/capitalone/Hygieia/issues/635

PR #648, opened 2016-06-23
Remove the default values for SONAR_USERNAME and SONAR_PASSWORD.

### LINKED ISSUE #647: Basic auth for sonarqube collector breaks unauthenticated behavior for sonarqube
On initial setup of the sonarqube collector, I received 401 errors every time the collector ran. My docker-compose.override.yml did not contain either `SONAR_USERNAME` or `SONAR_PASSWORD`. When I looked at the logs, those fields were defaulting to `sonarusername` and `sonarpassword`. After declaring `SONAR_USERNAME` and `SONAR_PASSWORD` as empty strings in my override file, they were still being overridden with the default value.

On closer inspection, the `sonar-codequality-collector/docker/sonar-codequality-properties-builder.sh` file sets `SONAR_USERNAME` to "sonarusername" using parameter substitution:

```
#Sonar Authentication Username - default is blank
sonar.username=${SONAR_USERNAME:-sonarusername}
```

There are two issues here. First of all, the default is clearly not blank - if `SONAR_USERNAME` is not provided, it will be set to "sonarusername". Second, this is using the wrong form of parameter substitution. `${FOO:-bar}` sets $FOO to "bar" if $FOO is undefined OR if $FOO is defined but null. This makes it impossible to set the username or password to an empty string.

As a result, there is no way to force an unauthenticated connection when using docker-compose, since the properties-builder file will always overwrite an empty value. The expected behavior should be to use the user-provided username/password without a default value. This way the username and password will only be set if provided by the user explicitly.

cbrenton:
Resolves #647.

### DIFF
--- a/collectors/build/sonar/docker/sonar-codequality-properties-builder.sh
+++ b/collectors/build/sonar/docker/sonar-codequality-properties-builder.sh
@@ -60,10 +60,10 @@ sonar.cron=${SONAR_CRON:-0 0/5 * * * *}
 sonar.servers[0]=${SONAR_URL:-http://localhost:9000}
 
 #Sonar Authentication Username - default is blank
-sonar.username=${SONAR_USERNAME:-sonarusername}
+sonar.username=$SONAR_USERNAME
 
 #Sonar Authentication Password - default is blank
-sonar.password=${SONAR_PASSWORD:-sonarpassword}
+sonar.password=$SONAR_PASSWORD
 
 #Sonar Metrics
 sonar.metrics=${SONAR_METRICS:-ncloc,line_coverage,violations,critical_violations,major_violations,blocker_violations,sqale_index,test_success_density,test_failures,test_errors,tests}


jdamick:
agreed, can you please fill out the CLA

tabladrum:
@cbrenton not sure if you are still around...

cbrenton:
@jdamick @tabladrum sorry, just filled it out.

mrpudn:
@cbrenton can you just resolve the branch conflicts :)

cbrenton:
@mrpudn I'm trying to, and they're resolved on my end, but I'm still seeing a conflict message here. Not sure what I'm doing wrong

mrpudn:
hmm, yeah it's still way behind the upstream repo's master branch.

What steps did you follow?

Usually what I do is have my `origin` remote point to my fork and my `upstream` remote point to the upstream capitalone/Hygieia repo, and then:
```git bash
git checkout master
git pull upstream master
git checkout <my-branch>
git rebase master
git push origin <my-branch>
```

cbrenton:
Ah, that's the problem. I was rebasing to my fork, not the upstream.

mrpudn:
@cbrenton can you make sure that the changes are what you intended - looks like a lot of unrelated changes

cbrenton:
@mrpudn fixed, thanks for the heads up

tabladrum:
LGTM

PR #663, opened 2016-06-29
Changes to Hygieia Jenkins Plugin- Support for Sonar 5.2+ and for new code metrics

rohans84:
## High Level Changes
1. Making the hygieia jenkins plugin work with Sonar 5.2+ versions. 
2. Capture new sonar metrics available from Sonar 5.0 onwards: new_violations, new_major_violations, new_critical_violations,new_blocker_violations, new_coverage, new_line_coverage, new_lines_to_cover

These new metrics capture metrics of only incremental code thus helping to fix "the leaky water pipe problem"

##  Background 
From Sonar 5.2+ onwards, SonarQube has hived off analysis of reports as an asynchronous process. This means that when Sonar analyss report is uploaded to Sonar during the sonar:sonar goal, the analysis does not get immediately completed, but Sonar sets an API endpoint called Compute Engine (CE) that gets updated when the corresponding analysis is available.

Hence, with Sonar 5.2+, the Jenkins job will need to monitor the CE API to determine if the processing is complete. After that is complete, can the Hygieia Publisher fetch the metrics from SonarQube for the project.

Also, the CE API can be queried only upto a maximum of 'X' attempts. This value is by default configured to 30 in SonarQube. For max attempts=30, Sonar recommends to make calls to CE API after every 10 seconds.

## Changes made

1. Added additional configuration by modifying HygieiaPublisher and by adding 2 html help files
    i. Max attempts to CE API
    ii. Interval period between successive attempts to CE API
 2. Modified SonarBuilder.java to:
    i) Determine CE API URL by inspecting the log from sonar analysis output

    ii) If CE API URL is found, it means the Sonar version is 5.2+. For this, keep polling CE API upto 'X' times with interval of 'Y' seconds. 

    iii) If CE API provides a SUCCESS response within maximum number of attempts, continue processing the Sonar resources API

    iv) If CE API does not provide a SUCCESS response within maximum number of attempts or provides a failure response, then do not invoke the Sonar Resources API as the report wasn't available. 

    v) If CE API URL is not found in the Sonar logs, it means the version is <5.2 and so the processing can continue as before.


tabladrum:
Can we put the polling number in the job config?

rohans84:
@tabladrum - The polling number (called max number of attempts is already added as a job config parameter for Hygieia along with the polling interval. The config items are ceQueryMaxAttempts and ceQueryIntervalInSeconds.

The html files were also added but I have added more information to the help file for easy understanding.

jimzucker:
@rohans84 Did you review docker/sonar-codequality-properties-builder.sh to ensure all properties generated for the container are in line with your changes?

rohans84:
@tabladrum - this changes is only for hygieia jenkins plugin. The Sonar collector already works in polling mode directly to the Sonar Resources API so change is not required in the sonar collector for sonar 5.3 compatibility - it will work out of the box.

However, the sonar collector needs to be changed to fetch new code metrics. But that feature is independent of this change and so I will create another pull request for it once I have completed the development.

hygieia-jenkins-plugin/src/main/java/hygieia/builder/SonarBuilder.java
```diff
@@ -33,8 +36,12 @@
...
     public static final String URL_PROJECT_ID_FRAGMENT = "/api/projects?format=json&key=%s";
-    public static final String URL_METRIC_FRAGMENT = "/api/resources?format=json&resource=%s&metrics=%s&includealerts=true";
-    public static final String METRICS = "security-violations,ncloc,violations,critical_violations,major_violations,blocker_violations,violations_density,tests,test_success_density,test_errors,test_failures,coverage,line_coverage,sqale_index";
+
+    public static final String CE_URL_PATTERN_IN_LOGS = ".*" + Pattern.quote("More about the report processing at ") + "(.*)";
+    public static final String CE_URL_PROJECT_ID_FRAGMENT = "/ce/task?id=%s";
+
+    public static final String URL_METRIC_FRAGMENT = "/api/resources?format=json&resource=%s&metrics=%s&includealerts=true&includetrends=true";
+    public static final String METRICS = "security-violations,ncloc,violations,critical_violations,major_violations,blocker_violations,violations_density,tests,test_success_density,test_errors,test_failures,coverage,line_coverage,sqale_index,new_violations,new_blocker_violations,new_critical_violations,new_major_violations,new_coverage,new_lines_to_cover,new_line_coverage";
```
tabladrum: Lets capture the trend too. Example: http://nemo.sonarsource.org/api/resources?resource=org.codehaus.sonar:sonar&metrics=new_coverage&includetrends=true

hygieia-jenkins-plugin/src/main/java/jenkins/plugins/hygieia/HygieiaPublisher.java
```diff
@@ -169,15 +170,47 @@ public boolean isPublishBuildStart() {
...
+        /** Sonar 5.2+ changes: get query interval from config
+         * If value is empty or null - return 10 (recommended value from SonarQube)
+         * @return max number of attempts to query Sonar CE API (10 if blank)
+         */
+        public String getCeQueryIntervalInSeconds() {
+            if (!StringUtils.isEmpty(ceQueryIntervalInSeconds)) {
+                return ceQueryIntervalInSeconds;
+            }
+            else {
+                return "10";
```
tabladrum: Declare as constant.

hygieia-jenkins-plugin/src/main/java/jenkins/plugins/hygieia/HygieiaPublisher.java
```diff
@@ -169,15 +170,47 @@ public boolean isPublishBuildStart() {
...
+        /** Sonar 5.2+ changes: get query max attempts from config
+         * If value is empty or null - return 30 (recommended value from SonarQube)
+         * @return max number of attempts to query Sonar CE API (30 if blank)
+         */
+        public String getCeQueryMaxAttempts() {
+            if (!StringUtils.isEmpty(ceQueryIntervalInSeconds)) {
+                return ceQueryMaxAttempts;
+            }
+            else {
+                return "30";
```
tabladrum: Declare constant

hygieia-jenkins-plugin/src/main/java/hygieia/builder/SonarBuilder.java
tabladrum: NVM. I see that you already do.

hygieia-jenkins-plugin/src/main/java/hygieia/builder/SonarBuilder.java
```diff
@@ -73,21 +83,123 @@ private void setSonarDetails(AbstractBuild build, String buildId) throws IOExcep
...
     public CodeQualityCreateRequest getSonarMetrics() throws ParseException {
         if (StringUtils.isEmpty(sonarServer) || StringUtils.isEmpty(sonarProjectID)) return null;
         String url = String.format(sonarServer + URL_METRIC_FRAGMENT, sonarProjectID, METRICS);
         RestCall restCall = new RestCall(publisher.getDescriptor().isUseProxy());
-        RestCall.RestCallResponse callResponse = restCall.makeRestCallGet(url);
-        int responseCode = callResponse.getResponseCode();
-        if (responseCode == HttpStatus.SC_OK) {
-            String resp = callResponse.getResponseString();
-            return buildQualityRequest(resp);
+
```
tabladrum: How does it work with the older version of Sonar?

rohans84: In the old version, there is no concept of Compute Engine. Sonar processes the report during the analysis part itself and hence when the mvn sonar:sonar completes, old Sonar would already have generated the report. Thus, the Hygieia plugin can access theh Resources API and have the results.

In the new version, when the sonar:sonar goal is completed, Sonar exposes an endpoint for the report that has a Status element. Its value changes to Success once the report is available.

hygieia-jenkins-plugin/src/main/java/jenkins/plugins/hygieia/HygieiaPublisher.java
rohans84: added constant and pushed the commit

hygieia-jenkins-plugin/src/main/java/jenkins/plugins/hygieia/HygieiaPublisher.java
rohans84: added constant and pushed the commit

Issue #667, opened 2016-07-01
Cleanup Eclipse PMD tests


jimzucker:
The # of PMD exceptions for routine things has grown, I am doing a cleanup so we can see real issues(if any)

- Private static vars should be capital
- removed unused imports
- removed unused variables

PR #668, opened 2016-07-01
#667 cleanup PMD warnings

jimzucker:
Cleaned up the basic ones, here is what's left
screenshot 2016-06-30 18 06 54

Here are the 'before errors'

screenshot 2016-07-01 08 03 03
screenshot 2016-07-01 08 02 50

### DIFF
--- a/api/src/main/java/com/capitalone/dashboard/service/FeatureServiceImpl.java
+++ b/api/src/main/java/com/capitalone/dashboard/service/FeatureServiceImpl.java
@@ -311,7 +311,7 @@ public class FeatureServiceImpl implements FeatureService {
 
 		default:
 			storyEstimates = new ArrayList<>();
-
+			break;
 		}
 
 		List<Feature> cumulativeEstimate = new ArrayList<>();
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/capitalone/dashboard/config/collector/CloudConfig.java
```diff
@@ -9,7 +9,7 @@
 
 public class CloudConfig extends CollectorItem {
     private static final String CLOUD_PROVIDER = "provider";
-    private static final List<NameValue> tags = new ArrayList<>();
+    private static final List<NameValue> TAGS = new ArrayList<>();
```
tabladrum: Is PMD complaining about "tags"?

api/src/main/java/com/capitalone/dashboard/service/FeatureServiceImpl.java
```diff
@@ -311,7 +311,7 @@ public FeatureServiceImpl(ComponentRepository componentRepository,
 
 		default:
 			storyEstimates = new ArrayList<>();
-
+			break;
```
tabladrum: +1

core/src/main/java/com/capitalone/dashboard/config/collector/CloudConfig.java
jimzucker: because it is static final PMD says it should be all CAPS

screenshot 2016-07-01 08 02 50

api/src/main/java/com/capitalone/dashboard/service/FeatureServiceImpl.java
jimzucker: see screenshot above, PMD complains in eclipse missing break, adding it is harmless and lest us see real issues in PMD.

PR #675, opened 2016-07-06
Jira Collector Performance, Robustness, and General Improvements

MarkRx:
Refactored, fixed, and improved code. Major performance improvements for bulk loads.
Before: load 7300 issues (page size 200) in 45.3 minutes
After: load 7300 issues (page size 200) in 2.45 minutes

Changes:
Performance:
- change delete + insert combos to upserts
- bulk saves instead of save one by one
- caching of epic data
- bulk epic pull at the end of each issues page
- reuse connections instead of creating one per each epic query

Improvements:
- fix improper Exception usage
- fix improper Spring usage
- simplified code
- improved logging
- added basic unit tests

Bug Fixes:
- fix issue where incorrect sprint is pulled when stories have multiple sprints
- fix NPE when story has a comma in the title
- fix issue where stories are lost if query collection is killed before it finishes (sort ASC order for pulling issues since our min date is based on the latest date found in the db)

Planned Changes (for a future PR):
- Add UI option to specify reporting on story points or hours
- Add configuration to specify hours feild
- Automatically obtain Todo/Doing/Done configurations

Issue #673, opened 2016-07-05
Confusion on Jira Time Tracking


MarkRx:
Hello,

The current jira implementation seems to pull time tracking data from the "timetracking" field in the REST json response for Stories. This seems strange as time tracking shouldn't be done on stories directly but instead on sub-tasks since stories are abstractions that use abstract measurements (story points). In our system stories do not have time tracking so that field is always empty. As a result the jira widget is always reporting 0 for todo/in progress/done. It does have some fields that appear to be aggregates of sub-task timings (like "aggregatetimespent" in seconds) but those only show up if sub tasks are actually created.

So questions:

1. Why does the current implementation pull time tracking information from stories?
2. Does someone have an example of a system that is currently using the jira collector?

If the current reporting behavior is intentional then I was thinking I would make a code change to allow users to configure the widget to report on either hours or story points. I'm not quite sure what fields to pull for hourly estimates though due to what I mentioned above.

tabladrum:
@MarkRx: full discloser - we have limited expertise on Jira. Sounds like you are deeper into Jira understanding than we are. Propose a PR and we can collaborate. THANKS!

jimzucker:
@MarkRx  @tabladrum JIRA will normally sum all the children onto the parent so pulling from there should be OK.  I think the salient point is you may want 'story points' vs hours, I would think most people want the 'Story Points' but some of us use hours. So maybe in the pull request to give a choice would be good..

MarkRx:
@tabladrum @jimzucker @amp5208 Is there a team/company using the jira collector? It looks like the estimates for stories are based on issue.getTimeTracking().getRemainingEstimateMinutes(). If that corresponds to the remaining number of minutes a task has then that seems wrong. The UI would incorrectly report the in-progress hours value and the "done" value as those would constantly be changing when work is completed. For example, if I have (10, 12, 5) (todo, inprogress, done) and a task in progress goes from 5 hours remaining to 3 hours remaining I now have (10, 10, 5) instead of (10, 10, 7).

MarkRx:
Here is what I might do:

For jira change sEstimate to use getOriginalEstimateMinutes(). Add a sStoryPoints field for story points and pull the value from a configurable field returned in the rest call. VersionOne will populate the sEstimate and sStoryPoints field with the same value. The UI will have a configuration for jira to specify a reporting metric of either hours or story points.

MarkRx:
[...]

PR #682, opened 2016-07-08
Jira allow configuration of estimation metric - story points or hours

MarkRx:
Allows each dashboard jira feature widget the ability to specify to report numbers based on story points or hours. Defaults to story points as that should be the most common use case. Configuration only present for the Jira feature collector configuration for now.

configjiraestimate

Implementation is a breaking change. sEstimate in Feature.java will be the story points value for all collectors (previously it was story points for versionone and hours for jira). A new field sEstimateTime will hold the time estimate (in minutes for precision) for jira. Jira now pulls hours estimates from the original estimate instead of remaining time due to issues described in #673.

Since I repurposed sEstimate for jira to be story points instead of hours (for consistency - the alternative was to make a bunch of one-off changes for a new sStoryPoints field) previous jira feature data will have to be repopulated for teams that use the jira collector. This can be done by dropping all feature documents. The jira collector will then repopulate everything the next time it runs. This should be relatively fast due to PR #675


tabladrum:
LGTM

MarkRx:
@tabladrum what do I need to do to get approved for this LGTM?

tabladrum:
More push to @jdamick and @amitmawkin :-) 
Guys: please check LGTM

PR #723, opened 2016-07-27
New Widget: AWS status

brostone51:
Added a new widget that can be added to the dashboard. The user enters a name and url, and then angular will query the URL, displaying an icon based on the result of the query.


tabladrum:
Cool. Can you add a screen shot (make sure nothing sensitive shows up :-)

amitmawkin:
also will this handle https ?

amitmawkin:
I will test the PR...

brostone51:
Yea, the two URL's I was testing with worked with https, so it should. I will attach a screen shot first thing tomorrow when I get into work.

brostone51:
@tabladrum @amitmawkin 
screen shot 2016-07-28 at 8 50 52 am
screen shot 2016-07-28 at 8 54 13 am
screen shot 2016-07-28 at 8 58 05 am

It's worth noting that the requested website needs to allow the Hygieia URL to make cross-domain requests to the endpoint.

MarkRx:
Is this limited to AWS? Or does it work on any generic REST URL?

tabladrum:
Can we make this for any http url? That would be awesome and that's what I had in my mind for long.

brostone51:
Yes it works for any http url, I suppose my name of "AWS Status" is unfortunate. As long as the requested URL allows [cross-origin requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS) (aka has a 'Access-Control-Allow-Origin: <hygieia-url>' header) then it should work. We are working on deploying kdc.capitalone.com right now, and hoping that this will trick the broswer into thinking our request to https://rwds-accts-cc-lkup-qa.kdc.capitalone.com/ext-rewards-accounts-cc-lookup-web/rewards/accounts/credit-card/health isn't cross-domain, will report back here with the results.

MarkRx:
To bypass the cross-origin issue you could create an API call that makes the REST calls from java instead of the cleint side browser.

tabladrum:
Here is my proposal: Call it "Monitoring" widget. Replace the current Monitoring with this. The purpose of this will be to monitor any http end point. Explore- Call the API to reach out to the endpoint assuming the API will have "access" to the URL.
Thoughts?

jimzucker:
+1

On Jul 28, 2016, at 9:58 AM, tabladrum <notifications@github.com<mailto:notifications@github.com>> wrote:


Here is my proposal: Call it "Monitoring" widget. Replace the current Monitoring with this. The purpose of this will be to monitor any http end point. Explore- Call the API to reach out to the endpoint assuming the API will have "access" to the URL.
Thoughts?

brostone51:
I have refactored the project to rename the new widget "monitor2". However, Java is not my strong suit, so I'm still having trouble getting the url call to route through the API. Will push to here once I get it working.

tabladrum:
Can you rebase? Thanks.

amitmawkin:
LGTM

tabladrum:
Lgtm

PR #742, opened 2016-08-09
Fixed gulp watch js files change

### LINKED ISSUE #741: UI - Gulp adding/removing js file does not update
After running gulp serve task and adding a new js controller it does not insert the new script upon refresh. Vice versa for removing a script. Modifying existing scripts works fine.

njgrisafi:
Adding and removing js files will update included scripts in index.html.

resolves #741


amitmawkin:
LGTM

tabladrum:
LGTM

PR #762, opened 2016-08-22
Fixed password match login error

### LINKED ISSUE #702: Login UI Bug
If you give an incorrect password the first time you are trying to login then even if you give the correct password second time, it  still doesn't login. If you refresh the page it starts working.

mrpudn:
Fixes #702 

- General refactoring of the `LoginController` and `login.html` template.
- Fixed validation error in `LoginController`:
  - Basically, this line:

    ```js
    $scope.lg.username.$setValidity('invalidUsernamePassword', data);
    ```
    This causes `lg.$valid` to always be false, which prevents the `loginData.login()` function from ever being called again.

### DIFF
--- a/UI/src/app/dashboard/controllers/login.js
+++ b/UI/src/app/dashboard/controllers/login.js
@@ -2,89 +2,50 @@
  * Controller for performing authentication or signingup a new user */
 (function () {
     'use strict';
-
-    angular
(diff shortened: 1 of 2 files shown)


tabladrum:
LGTM

amitmawkin:
LGTM

PR #798, opened 2016-09-07
Fix property generator scripts to generate 'dbname' property instead of 'database'

MarkRx:
Almost all of the property generator scripts use a property named 'database' to specify what database to use. The code uses 'dbname'. Someone probably changed this property a while ago and forgot to update the property generator scripts. It has gone unnoticed because people probably have just used the default value for the database name.

This change fixes the property generator and property files to properly use 'dbname'


amitmawkin:
LGTM

tabladrum:
LGTM

tabladrum:
@MarkRx needs rebase :-(

MarkRx:
branch updated

PR #824, opened 2016-09-15
Add scm parent revision numbers to Commits

MarkRx:
Just as the title says. It will prove to be useful information in the future for some things I have planned. I wanted to get it in now so collectors can start filling up with the data.

Note that the parent revision data will only come from the commit collectors. SCM objects created by build collectors will not have the parent revision information (as it doesn't seem to be available in the api). Hence at present commits in the pipeline document for build and environments will not contain the parent scm revision number information.


MarkRx:
rebased

amitmawkin:
LGTM

MarkRx:
@tabladrum I think you also have to say LGTM as approving the request doesn't give the LGTM approval

tabladrum:
LGTM

Issue #599, opened 2016-05-27
Incorrect javascript ordering on index.html


MarkRx:
Hello,

I'm having issues with angular module ordering. Not sure exactly how to reproduce this. It happens after starting up gulp again. Every now and then the index.html  portion will get out of order resulting in a bunch of errors:

> 
> bower.js:1470 Uncaught Error: [$injector:nomod] Module 'hygieia-dashboard.core' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.
> http://errors.angularjs.org/1.3.20/$injector/nomod?p0=hygieia-dashboard.core(anonymous function) 
> @ bower.js:1470(anonymous function) 
> @ bower.js:3185ensure 
> @ bower.js:3109module 
> @ bower.js:3183(anonymous function) 
> @ commit-blocks.js:2

Here is the order when it does not work. Observe that widgets are loaded before core/module.js.

```
<script src="app/app.js"></script>
<script src="components/templates/capone.js"></script>
<script src="components/widgets/build/config.js"></script>
<script src="components/widgets/build/detail.js"></script>
<script src="components/widgets/build/module.js"></script>
<script src="components/widgets/build/view.js"></script>
<script src="app/chartist/plugins/axisLabels.js"></script>
<script src="app/chartist/plugins/axisTitle.js"></script>
<script src="app/chartist/plugins/gridBoundaries.js"></script>
<script src="app/chartist/plugins/lineAboveArea.js"></script>
<script src="app/chartist/plugins/pointClick.js"></script>
<script src="app/chartist/plugins/pointHalo.js"></script>
<script src="app/chartist/plugins/pointlabels.js"></script>
<script src="app/chartist/plugins/threshold.js"></script>
<script src="app/chartist/plugins/tooltip.js"></script>
<script src="components/widgets/chatops/config.js"></script>
<script src="components/widgets/chatops/module.js"></script>
<script src="components/widgets/chatops/view.js"></script>
<script src="components/widgets/pipeline/config.js"></script>
<script src="components/widgets/pipeline/module.js"></script>
<script src="components/widgets/pipeline/view.js"></script>
<script src="components/widgets/deploy/config.js"></script>
<script src="components/widgets/deploy/detail.js"></script>
<script src="components/widgets/deploy/module.js"></script>
<script src="components/widgets/deploy/view.js"></script>
<script src="components/widgets/product/module.js"></script>
<script src="components/widgets/product/view.js"></script>
[...]

PR #825, opened 2016-09-15
Add gulp-order to enforce index.html js import order

MarkRx:
Fix for #599

Currently the ordering of the hygieia:js files in index.html is non deterministic. It seems to change on me when file timestamps change. This PR will force <script src="app/dashboard/core/module.js"></script> to be at the top.


tabladrum:
LGTM

PR #854, opened 2016-09-21
Added template for standalone cloud widget

amitmawkin:
1. Code change for showing all the tabs moment the dashboard is created.
2. Ability to create standalone AWS cloud dashboard, This tab was earlier only visible if deploy widget was configured.
3. Modified Maintainers file to add new members

image

image

### DIFF
--- a/UI/src/components/templates/capone.html
+++ b/UI/src/components/templates/capone.html
@@ -7,9 +7,7 @@
 
                 {{::ctrl.dashboard.title}}
 
-                <div class="pipeline-toggle btn-group"
-
-                     ng-show="template.hasComponents(ctrl.dashboard, ['SCM'])">
+                <div class="pipeline-toggle btn-group">
 
                     <span class="btn" ng-repeat="tab in template.tabs" ng-click="template.toggleView($index)"
                           ng-class="tab.name == template.widgetView ? 'btn-primary' : 'btn-default'">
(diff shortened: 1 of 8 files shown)


UI/src/components/templates/capone.html
```diff
@@ -9,7 +9,7 @@
 
                 <div class="pipeline-toggle btn-group"
 
-                     ng-show="template.hasComponents(ctrl.dashboard, ['Deployment'])">
+                     ng-show="template.hasComponents(ctrl.dashboard, ['SCM'])">
```
tabladrum: What does this do again?

amitmawkin: currently the pipeline and cloud widget are only shown when somebody configures the Deployment widget, this will allow the tabs to be shown on the upper right if SCM widget is configured. I don't think this is needed any more as i created a separate cloud widget

tabladrum: Yeah. I think the ng-show is really not needed.

amitmawkin: You mean this should be visible if just somebody creates a Dashboard, BTW this is how it will look

image

amitmawkin:
LGTM

amitmawkin:
LGTM

tabladrum:
LGTM. :+1:

PR #857, opened 2016-09-22
Change jenkins collector to pull scm branch information

MarkRx:
- Have jenkins pull branch information if available


tabladrum:
There are few ways this might fail. One such can be this situation:
1. Setup Jenkins job with two git repo (not using MultiSCM plugin). Commit to both and you will notice that there are two back to back Jenkins builds - for each repo changes. See if your code can match the right branch.
2. If there are multiple commits in a single build, the branch refers to the top level commit. In a build with changes to multiple git repos (using Multi SCM plugin), there is no simple way to match the branch - unless we assume certain formatting to be true: origin refers to repo[0], origin1 refers to repo[1] etc.

MarkRx:
@pxk5958 could you take a look?

pxk5958:
@tabladrum Thanks. I am looking into the various cases.

MarkRx:
What we are going to do is given the branch information and the originX number we can determine the remoteUrl by pulling the job config that was used for the build and correlating the X to the X'th repo. 

It turns out that the "remoteUrls" list is buggy to begin with so this should be more reliable. I found a case where I specified multiple repositories but it ended up building one of them and claiming it built the other one.

pxk5958:
There are 2 cases to consider here.

1. Multiple repositories configured through Jenkins MultiSCM plugin. 
In this case, each remoteUrl is in a separate action in the json response, so we just have to iterate over the actions and find the remoteUrl and corresponding branch in the lastBuiltRevision of each action. 

2. Multiple repositories configured through the 'Add repository' button on the Jenkins Git plugin.
In this case, the remoteUrls are together in the same action and are not necessarily in the correct order to associate the remote origin, origin1, origin2,...,origin[n] with the corresponding url. There are 2 ways to go about solving this:
    (a) Parse the jenkins job config.xml and find out the remoteUrls in order. There are 2 drawbacks to this approach. If the config.xml file was changed in the time between when a job is triggered and before the queued job finally executes, we cannot get the right config.xml since we don't have the information as to when exactly the job was triggered (as opposed to when it was executed which is available in the json). The other drawback is that if the repository url is parameterized in the config, we won't have the correct url as well.
    (b) Jenkins Git plugin currently stores the remoteUrls in a HashSet which is why they are unordered. I modified it to a LinkedHashSet so that we can get them ordered and hence associate with the correct branch. I have raised an issue (https://issues.jenkins-ci.org/browse/JENKINS-38800) and have submitted a pull request with this change (https://github.com/jenkinsci/git-plugin/pull/441) but the maintainer doesn't see so far why we would need to configure multiple repositories.

Also, like @MarkRx mentioned, sometimes the build information is wrong when multiple repositories are configured in the Git plugin, but it happens only if we change the job configuration by adding more repositories after at least 1 build has already been done. In other words, results are consistent as long as we never add more repositories after a job has been created and executed at least once.

Considering these, I think we should drop support for multiple repositories in Git plugin since it seems buggy but should support adding branch information for single repositories and MultiSCM plugin. What do you think? @tabladrum

amitmawkin:
@tabladrum lets not support multi repository, nothing good comes out of it.

MarkRx:
I agree with @amitmawkin. I've conversed a lot with @pxk5958 and we came to the same conclusion. The "multi-repository" functionality for the git plugin is broken and unreliable. Even if it worked as intended its use case is questionable at best. The maintainers agree. They suggest using the multi-scm plugin for multiple git repos. The code changes @pxk5958 will work for multi-repository.

amitmawkin:
LGTM

tabladrum:
Ok. I am good with not supporting multi repository. Can we think of a way to put a big bold disclaimer or something?

tabladrum:
LGTM

MarkRx:
r

PR #860, opened 2016-09-22
Change will generate gzip of UI project dist folder

amitmawkin:
Used Maven assembly plugin to assemble UI project dist folder in tar gzip format
This will allow folks to consume each build artifact from Maven Central, with out need to build from source


tabladrum:
LGTM

amitmawkin:
LGTM

Issue #812, opened 2016-09-13
Prevent Creation of Dashboard with same name.


amitmawkin:
Different users can create dashboard with same name. The System should prevent that and api should return an error code(400 Bad Request) with helpful message.

amitmawkin:
Fixed with https://github.com/capitalone/Hygieia/pull/853

PR #868, opened 2016-09-25
Add troubleshooting section to readme to address dashboard dup titles.

satishc1:
Add troubleshooting section to readme to address any existing duplicate dashboard titles.

https://github.com/capitalone/Hygieia/issues/812 was fixed by adding a unique index to the title field in the dashboards collection. 

But If there happen to be any duplicate titles already existing in the database, the api module will fail to launch with the following error :
Error creating bean with name 'dashboardRepository': Invocation of init method failed; nested exception is org.springframework.dao.DuplicateKeyException: Write failed with error code 11000 and error message 'null'.

This problem can be fixed by running the fixdups.js file on the mongodb instance as mentioned in the readme.


tabladrum:
LGTM

satishc1:
LGTM

PR #873, opened 2016-09-27
docker-compose syntax and linking upgraded to v2, mongo setup automated

charleswisniewski:
@tabladrum impetus behind these changes:
- bring docker compose syntax up to date
- speed up the first time set up of Hygieia - quicker for onboarding

Thoughts?


tabladrum:
@jimzucker @charleswisniewski like it! 
@mrpudn want to take it? :-)

jimzucker:
@tabladrum a couple of points after reviewing the code for your consideration:

1. A minor point I think the sql file should be .sql not .js as .js is normally javascript.

2. I haven't had time to run a test, but reviewing the files db/db-setup.js & db/db-setup.sh I think we will get errors the second time it runs if you re-create everything as the Database will already exist.  I think there needs to be an if not exists user around the create commands. I think you need to do something like this: 

> mongo --port 27037 --quiet -u superuser -p pwd 
>     --eval "db.system.users.find({user:'db'}).count()" admin
> 

If it returns zero then the id does not exist and you should exec the id creation. (see http://stackoverflow.com/questions/12856420/mongodb-command-line-to-show-if-a-user-exists-for-puppet-unless-clause)

tabladrum:
@jimzucker Mongo "sql" are all js. 
LGTM

tabladrum:
LGTM

mrpudn:
@charleswisniewski can you resolve the merge conflicts? :)

xiaods:
need rebase and clean the code.

xiaods:
need update?

satishc1:
@tabladrum Merging this one. There are a few docker related PRs that will be merged on top of this.

kwhickey:
@satishc1 
This was merged in on July 6th, which gratefully upgraded the `docker-compose.yml` file to versions 2. 
But then #1275 was merged in right overtop of this, which reverted the compose syntax back to v1. I'm guessing that was unintentional and a bad merge sequencing, and the merge needs to be reapplied. 
See history for July 6, 2017:
https://github.com/capitalone/Hygieia/commits/master/docker-compose.yml

PR #879, opened 2016-09-29
Bug Fix for Hygieia App version display

amitmawkin:
Spring boot tokenization works only for properties bundles inside the spring boot jar. To cater to this and preserve backwards compatibility.
- Introduced a new file called version.properties
- File gets updated by maven during build phase with current version.
- The property value is injected in the class requiring this.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -32,6 +32,22 @@
 
 
     <build>
+        <resources>
+            <resource>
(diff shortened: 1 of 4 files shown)


amitmawkin:
LGTM

tabladrum:
LGTM

Issue #869, opened 2016-09-26
Bug: pipeline dashboard build commit time statistics off


MarkRx:
BuildEventListener.java is using build.getTimestamp() for the PipelineCommit timestamps. The timestamp on builds is being stamped with the system time of when the collector was run. Hence when the collector first runs and picks up a lot of builds these builds will have nearly identical timestamps. The pipeline will then report numbers based on these timestamps (when the collector ran as opposed to when the build ran).

The code should instead reference build.getStartTime() (or build.getEndTime()) to more accurately represent when the commit entered that stage of the pipeline.

mrpudn:
@MarkRx can we close this one? It appears to be resolved with #888

MarkRx:
yup

Issue #864, opened 2016-09-23
Product dashboard doesn't display data right away


MarkRx:
Not sure why yet but when I first load the product dashboard everything is blank. To get data to start showing I have to click the configure button to open that dialog then close it. Data will also show up if I wait a minute or so (perhaps some auto refresh gets triggered).

![image](https://cloud.githubusercontent.com/assets/17504314/18785319/981a330a-8165-11e6-8d09-ce0536534002.png)

Chrome Version 53.0.2785.116 m (64-bit)

sanjeevnewt:
I too not able to see the data even after many refresh.

MarkRx:
I'm not a web developer but I think some kind of refresh/notification needs to happen to the row after data is updated in product/view.js setTeamData. I suspect there is a periodic refresh of 1 minute somewhere that is causing the data to get updated but delayed by 1 refresh (so the second refresh shows the data from the 1st call, the 3rd refresh shows the data from the second call, etc).

tabladrum:
I will take a look at it tonight

sanjeevnewt:
@tabladrum any update on this one?

tabladrum:
It does take a few minutes to get the data for the first time. From that time onwards it uses the browser cache. So depending upon how much data you already have it may take about 2 minutes for the initial load.

MarkRx:
@tabladrum it still occurs even if the data response is very fast. I watched the response through the debugger and the data comes back in less than a second.

tabladrum:
Interesting.Let me hit it with more debugs.

MarkRx:
@tabladrum I think I found and fixed the issue. #1021

nireeshT:
Fixed with #1020

Issue #828, opened 2016-09-15
Bug: Commits in the pipeline only show up in the last configured environment


MarkRx:
Documenting this so it doesn't get forgotten about / lost...

This is due to the EventComponentEventListener being triggered on any environment_component that is saved. The XLD collector is resaving environment_components even if they have not changed. Not sure if the UDeploy collector is doing this too. As a result commits are being added to every environment in a pipeline which causes hygieia to interpret them as always being promoted to the last environment.

Code:

```
                if (!compList.isEmpty()) {
                    List<EnvironmentComponent> existingComponents = envComponentRepository
                            .findByCollectorItemIdAndEnvironmentID(application.getId(), environmentId);
                    envComponentRepository.delete(existingComponents);
                    envComponentRepository.save(compList);
                }
```

nireeshT:
@MarkRx  Update to latest code base and try with new plugin

Issue #816, opened 2016-09-13
Deploying old applications does not properly reflect in the product dashboard


MarkRx:
If an older version of a component is deployed it will not be reflected properly in the product dashboard. This is because the EnvironmentComponentEventListener always adds the commits from the builds from the latest artifacts. There is no logic to remove commits. Thus if an older version is deployed into an environment newer commits will still exist in the pipeline document resulting in an inaccurate dashboard.

MarkRx:
In order to fix this I think the following has to happen:

1. Artifacts need to store information about the scmRevisionNumber, scmUrl, and scmBranch that was used to create them. For a MVP we should assume an artifact comes from one repository.
2. Commits need to store information to allow a graph to be built in memory. A commit can store its parent commits (more than one for merges). A graph can be built in memory based on the HEAD information from the artifact. For scm we might be able to assume all scm revision numbers less than the one stored on the artifact are ancestors.
3. With the information from 1 and 2 determine the commits for the artifact being deployed in the environment in EnvironmentComponentEventListener. We would have to correlate the component version with the artifact version for now until a more flexible association mechanism can be devised.

tabladrum:
Will work on it this week.

MarkRx:
@tabladrum what do you plan on doing exactly so I don't bump in to your changes? Also I suggest working off of PR #815 for storing information in artifacts

tabladrum:
Let me first digest your analysis first. Are you open for a conf call tomorrow or Friday?

MarkRx:
Yes I am.

MarkRx:
PR #824 for part of point 2. Even if this solution isn't used I think having the scm parent revision information will prove to be useful in the future.

tabladrum:
Please send me your contact info to my work email: tapabrata.pal@capitalone.com. I will put something on the calendar.

PR #888, opened 2016-10-05
Dynamic Product Dashboard POC

MarkRx:
This PR adds a dynamic product dashboard pipeline service that can be enabled with the feature.dynamicPipeline=enabled feature switch. This requires PR #815. More details in comment https://github.com/capitalone/Hygieia/issues/804#issuecomment-247391585.

Basically data for commits, builds, deployments, and artifacts are associated dynamically using constructed composite keys. That correlated data is calculated on-demand. Doing so solves a number of synchronization issues that are pointed out below.

Data is correlated in the following ways: 
- Build -> Commit: Builds keep track of the SCM revision numbers as well as the repository and branch information (though builds with multiple repositories are not 100% accurate). Given a list of commits for the dashboard we can correlate builds to them using the scm revision number.
- EnvironmentComponent -> BinaryArtifact: Given a list of Environments we can gather DeploymentUnits and associate them to BinaryArtifacts by the component name and version number. In the future this information may be stored in metadata that is retrieved by the deployment collector. 
- BinaryArtifact -> Commit: An artifact will contain information about the HEAD svn revision number that was used to produce it. Given the scm revision number we can find it in our list of commits that are tracked for the dashboard and determine it along with all previous commits. For GIT this is done using a graph built from Commit.getScmParentRevisionNumbers(). For SVN we simply grab all revisions with a number less than ours.

This addresses the following issues
- #816 - Deploying old applications does not properly reflect in the product dashboard: only commits that come prior to the head commit for the version that was deployed are used
- #828 - Bug: Commits in the pipeline only show up in the last configured environment: since the EnvironmentComponent event listener doesn't play in to this this bug doesn't get hit
- #804 - Pipeline document not preemptively filled when created: pipeline is generated with each call
- #810 - NPE in CommitEventListener (bitbucket collector): partially fixed while I was in the code (no NPE)
- #869 - Bug: pipeline dashboard build commit time statistics off: fixed this while I was in the code
- #806 - Publishing artifacts requires build information to have already been collected: Fixed by PR #815

Sample artifact document:

```
{
    "_id" : ObjectId("57f2680746e0fb000e3026fc"),
    "_class" : "com.capitalone.dashboard.model.BinaryArtifact",
    "timestamp" : NumberLong(1475504135011),
    "artifactName" : "Helloworld",
    "canonicalName" : "Helloworld",
    "artifactGroupId" : "com.paychex.helloworld",
    "artifactVersion" : "4.9.7.20161003-101416I",
    "metadata" : {
        "jobName" : "mrichar2_helloworld",
        "scmUrl" : "https://repo.com/scm/~username/helloworld.git",
        "scmBranch" : "master",
        "buildUrl" : "http://jenkins.instance.com/job/helloworld/235/",
        "jobUrl" : "http://jenkins.instance.com/job/helloworld/",
        "buildNumber" : "235",
        "scmRevisionNumber" : "bea69104639e87faff235827e036e5021d1f93cf",
        "instanceUrl" : "http://jenkins.instance.com/"
    }
}
```

Note the following limitations / items not implemented at this time:
- Dynamic pipeline does not currently support SVN
- Date bounding is ignored as there is a current defect in the UI that shows itself here
- Only handles 1 component (though this is the same with the existing pipeline implementation)
- Jenkins branches are not 100% reliable - #857 - code currently does not filter on branch. We can get away with it for now due to commits already being filtered by branch. See comment in DynamicPipelineServiceImpl.filterBuilds for more info.
- Data doesn't display right away. This is an existing issue - #864. To work around open and close the options panel
- Jenkins commits sometimes get missed by builds. This is existing and occours with the event listener implementation as well. To work around it older commits that get skipped are added back using our best guess for a timestamp. This is a similar approach to what the BuildEventListener does
- Performance for very large datasets is not known. I don't think it will be a problem since it should be about on par with the amount of time to load team dashboards since it uses the same service calls
- Branch equality needs to be made smarter. Jenkins will use 'origin/master' whereas the user might only use 'master'


MarkRx:
rebased

MarkRx:
Fixed exception when git scm is specified with ssh://.

tabladrum:
LGTM

satishc1:
LGTM

Issue #678, opened 2016-07-07
Jira features without sprints default to KANBAN


MarkRx:
PR #606 added kanban support to the feature collector widget. For jira it looks like stories are defaulting to be kanban if they do not have a sprint associated to them. As a result stories that are on the backlog for teams that exclusively use scums are showing up as being part of a kanban sprint.

Not sure how to properly address the issue.

tabladrum:
@MarkRx Noticed that too. The assumption may not be correct for everyone. Let me do some digging in.

MarkRx:
Adding @amp5208 

I don't know if there is a way to programmatically determine the sprint type. In our system we have teams that use kanban by using really long single springs. There are other kanban teams that don't have any sprints period.

Not having a sprint cannot default to being kanban since in the context of a scrum team it means the story is on the backlog.

Perhaps the UI could be configurable.
- Kanban - all features for active sprints and stories without a sprint. Only show Total + WIP
- Scrum - all features for active sprints. Show total + WIP + Done
- Both - not sure this is a real use case. If it was there is overlap in the data above so not sure how it would work.

amp5208:
@MarkRx :  I have reviewed this issue and the open issued tied to it for context.  I used the logic of sprint length to determine if a team was kanban or scrum, based on the assumption you mentioned above.  I am aware that this is not perfect logic; however, Jira does not provide a programatic way or built-in field that I can with 100% certainty confirm if a Jira board is, in fact, kanban or scrum.  Their API does not support this, based on my research.

As for your suggestion, the "both" use case is already covered.  If a Jira project has both Kanban and Scrum stories, the UI will switch between all available Scrum and Kanban boards within the Jira project.  Let me know if you need additional clarification.

MarkRx:
Working on this now.

MarkRx:
@amp5208 do you have teams that have both active scrum and active kanban sprints? Is the 'both' option needed?

I'm looking at changing the sprint type to be a configuration item that is set. As such it doesn't make sense to me for the 'both' case to remove sprint types if no iterations of the type are found. If one type doesn't exist but the user explicitly set it to 'both' then I would expect to see both regardless of if there is data or not.

Also in the collector I have to remove the fake KANBAN sprint data that is being attached to stories if no sprints are found. This is because not having a sprint does not imply KANBAN - it could simply be a backlog item. Instead of trying to deduce scrum/kanban it would be configured in the UI.

MarkRx:
[...]

PR #903, opened 2016-10-18
Fix for feature widget not correctly handling some sprint type use cases - #678

MarkRx:
Fix for #678. Backwards compatible. Adds the ability to configure a widget to interpret sprints as either kanban, scrum, or both. This is necessary since it is not possible to deduce which sprint type a project uses from the rest api. I also did some code cleanup while there. The basic issue with the current code is that it interpreted features without sprints as being KANBAN which is not true in all use cases.

Here is how it works:
In order to be considered part of the sprint the feature must not be deleted and must have an "active" sprint asset state if the sprint is set. The following logic also applies:

A feature is part of a scrum sprint if any of the following are true:
- the feature has a sprint set that has start <= now <= end and end < EOT (9999-12-31T59:59:59.999999)

A feature is part of a kanban sprint if any of the following are true:
- the feature does not have a sprint set
- the feature has a sprint set that does not have an end date
- the feature has a sprint set that has an end date >= EOT (9999-12-31T59:59:59.999999)

Note the above are mutually exclusive.

Observe how this handles all of the following use cases:
- Typical scrums (a sprint is live for X number of weeks)
- Kanban without any sprints assigned
- Kanban using a sprint without an end date
- Kanban that was set from previous code that interpretted not having a sprint set as being KANBAN (since we use EOT in the queries above)
- Scrum & Kanban (don't see how this is a real world use case but it works similar to the previous code. Kanban will interpret backlog items as being part of kanban but that issue is existing and cannot be resolved due to lack of information)
- The code in the versionone widget that interprets sprints that have an end date > some cutoff as being KANBAN. This works as it will not end up setting the sprint data which causes the feature to be interpreted as kanban

@amp5208 @tabladrum


MarkRx:
rebased

tabladrum:
@MarkRx not sure how this conflict came in. Can you please check?

MarkRx:
rebased to keep history clean

tabladrum:
LGTM

satishc1:
LGTM

amitmawkin:
LGTM

PR #912, opened 2016-10-25
Allow skipping of frontend maven plugin

mdskd:
The frontend-maven-plugin downloads node from an Internet repository and installs for execution. This may not be feasible for all consumers.

This merge request allows configuration of skipping the frontend-maven-plugin, defaulting to false (in other words, it executes by default).


mdskd:
Did not mean to do this : )

PR #891, opened 2016-10-07
Handling multiple cross functional deployment servers

MarkRx:
This change allows the deployment widgets to handle more than one server. It has a configuration that can be set to allow this.

Use case: We have multiple deployment servers that are split up by function. One handles prod and prod-like environments. The other handles test, qa, and dev environments. Data gathered from multiple servers needs to be correlated together and treated as one application.

Changes:
- API now supports multiple deployment collectors per a single dashboard
- UI has ability to correlate applications together by ID (when enabled) thus allowing multiple deployment collector items to be used under the hood.
- Add API + UI configurable options ( #890 )

Changes will work with existing documents in the database.

PR #921, opened 2016-10-31
Add ability to ignore failures on certain environments

MarkRx:
Adds ability to configure a regular expression to indicate environments that can ignore errors. If an environment ignores errors the UI will use a gray icon instead of a red one. The deploy widget will only turn red if an environment that is not ignoring errors fails.

Use case: There are cases when we don't care if an environment is down. In a blue/green stage-active deployment we never care about the "staging" environment. Without this change the deploy widget will almost always be red (as usually no code is staged).

** Update **
Configuration is specified in the deploy widget config

Requires PR #891 

taxsdlc666-1
taxsdlc-666-2

Update: Added additional ability to configure on a per dashboard basis.
taxsdlc666-4

### DIFF
--- a/UI/src/app/dashboard/directives/dash-status.js
+++ b/UI/src/app/dashboard/directives/dash-status.js
@@ -9,7 +9,8 @@
 
         // status constant so widgets can use the same values as an enum
         .constant('DashStatus', {
-            PASS: 1,
+            IGNORE: 0,
+        	PASS: 1,
             WARN: 2,
             FAIL: 3
         })
@@ -20,7 +21,8 @@
         return {
             scope: {
                 status: '@dashStatus',
-                failText: '@dashStatusFailText'
+                failText: '@dashStatusFailText',
+                ignoreText: '@dashStatusIgnoreText'
             },
             restrict: 'A',
             controller: controller,
@@ -34,6 +36,7 @@
 
         function link(scope, element, attrs, containerCtrl) {
             scope.failText = scope.failText || '!';
+            scope.ignoreText = scope.ignoreText || '-';
 
             attrs.$observe('dashStatus', function() {
                 // accept a bunch of different statuses
@@ -54,6 +57,11 @@
                     case '2':
                     case 'warning':
                         scope.currentStatus = DashStatus.WARN;
+                        break;
+                    case 0:
+                    case '0':
+                    case 'ignore':
+                        scope.currentStatus = DashStatus.IGNORE;
                         break;
                     default:
                         break;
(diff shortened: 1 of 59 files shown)


tabladrum:
@MarkRx we might have merge some PR's out of order. Can you see if this conflict can be resolved?

tabladrum:
@MarkRx this needs few quick manual merging.

MarkRx:
rebased & squashed

tabladrum:
Looks like you are saving this config as a system global config. Better if we save it within the widget config. And then modify the deploy widget to have the user to enter the regex pattern. We don't want the same regex affecting every user or we don't want to change the properties file everytime there is a new regex.

Thoughts?

MarkRx:
Will look at next week.

MarkRx:
@tabladrum done. Screenshot above.

Rebased.

MarkRx:
@pxk5958

MarkRx:
Done. Will go through an rebase everything once #1042 is merged in.

MarkRx:
rebased onto master 2.0.5

UI/src/app/dashboard/directives/dash-status.js
```diff
@@ -55,6 +58,11 @@
                     case 'warning':
                         scope.currentStatus = DashStatus.WARN;
                         break;
+                    case 0:
+                    case '0':
```
tabladrum: what are these 0 and '0' status?

UI/src/components/widgets/deploy/module.js
```diff
@@ -21,6 +21,29 @@
 
     angular
         .module(HygieiaConfig.module)
+        .directive('validregex', function() {
+          return {
+            require: 'ngModel',
+            link: function(scope, elm, attrs, ctrl) {
+              ctrl.$validators.validregex = function(modelValue, viewValue) {
+                if (ctrl.$isEmpty(modelValue)) {
```
tabladrum: Do we need this check? I think the RegExp(..) will handle the empty scenario too?

nireeshT:
@MarkRx @tabladrum  - Does the regex to ignore failures check for environment names and deployment job names.

tabladrum:
@nireeshT just the environment names.

UI/src/app/dashboard/directives/dash-status.js
nireeshT: @tabladrum  - 0 or '0' status is newly added IGNORE: 0

tabladrum: Now I see it. Sorry.

PR #949, opened 2016-11-15
Allow jira to pick up on multiple issue types

MarkRx:
At present the jira collector can only collect on one issue type. We want to be able to pick up estimates assigned to other types (such as investigation spikes and defects). This change allows multiple issue types to be specified in the config.

Note the name of the config item was corrected so configs will have to be updated.

@pxk5958


MarkRx:
Added as part of #974

PR #958, opened 2016-11-17
Ldap integration2

brennanananan:
Currently the Hygieia product allows users to create accounts, and then stores the passwords to those accounts in Mongo. This PR adds a new authentication service that can be used for LDAP based authentication to gain access to Hygieia. It uses an existing LDAP server implemeted by the deploying organization to check if an account merely exists. It has not been tested against group level access, but by modifying the dn in the application.properties file, more fine-grain restrictions should be able to be enforced (see README).

LDAP authentication can be activated through the "ldapAuth" spring profile, and setup instructions have been added to the api README.


brennanananan:
Tests all pass when ran locally against the ldapIntegration2 branch after merging in master changes. Please advise.

MarkRx:
How do you login as an admin?

How does one turn this on or off using configuration in the docker files and docker-compose?

brennanananan:
I was unaware of there being admin functionality without these changes.
To run with LDAP, the Dockerfile CMD for the API would be changed to say
```java -jar ....... --spring.profiles.active=ldapAuth``` 

Running without LDAP is the default behavior.

tabladrum:
@brennanananan build failed. I manually restarted it. Let's see.

tabladrum:
@brennanananan build passed. magic! :-)

tabladrum:
@amitmawkin @satishc1 @MarkRx for review.

amitmawkin:
@brennanananan undocumented feature, if you create a user with name admin you get to see an extra gear on the right top hand corner  that allows you to delete any dashboard and set themes.

brennanananan:
Thanks @amitmawkin , I will take a look at how to incorporate that feature into my changes. Does the admin 'group' only get those two extra actions?

pxk5958:
Also, it seems you're assuming the usernames will be stored under uid. It may not be so and needs to be configurable as a search filter.

MarkRx:
- You need to change the docker properties builder so that no rebuild is necessary to be able to use LDAP
- The "sign up" feature needs to be hidden if ldap is enabled
- The "sign up" api call needs to throw an exception if ldap is enabled
- I would prefer reusing the existing auth controller and just checking a property dynamically to see if LDAP is enabled instead of using a spring profile. You can stick it into the systemConfig property map for now that is part of API. In the future we may want to build a System Config UI that can change these values and store them into the database. Using spring profiles prevents dynamic property changes without a restart.

tabladrum:
@brennanananan any update on @MarkRx comments and the admin user? Let me know if you need extra hands.

brennanananan:
@tabladrum I agree with the comments that @MarkRx supplied, however I currently do not have the bandwidth to work on them. It would be possible to change the authentication service dynamically based on the user trying to authenticate, but it might make more sense to hold off on the ldap integration until someone is able to implement role based access on top of what I have provided.

MarkRx:
@brennanananan @pxk5958 can pick up the work if you don't have the time

pxk5958:
@brennanananan Any update on this?

brennanananan:
@pxk5958 I have not been able to look further into this at this time. I may be able to start looking into this again starting next week

brennanananan:
Closing as this requires more thought and design, see #1031

Issue #969, opened 2016-11-22
Encrypted Properties


Mpluya:
Our organization has a requirement to not store credentials (i.e. database credentials) in plaintext on the property files.
I noticed the codebase already has support for encrypt/decrypt by using the Encryption class on the core project.
The only downside in my opinion is I now have to add code to call the encryption class. It would be nice if my encypted properties are served up automatically decrypted.
That said, I found jasypt-spring-boot (Jasypt integration for Spring boot). It boot straps a Spring post processor which makes the app aware of the encrypted properties. Details about jasypt-spring-boot is here https://github.com/ulisesbocchio/jasypt-spring-boot.

My question is, would there be any interest in incorporating this in Hygieia?

tabladrum:
@Mpluya Thanks for the link. Definitely interested.

Mpluya:
@tabladrum cool. Thank you. I'll submit a PR early next week at the latest.

satishc1:
Closed.

PR #991, opened 2016-11-29
Property encrypt

Mpluya:
This is related to the issue (question) I asked a week ago -- https://github.com/capitalone/Hygieia/issues/969

This enhancement just brings in jasypt-spring-boot and samples on encrypting the database password.


MarkRx:
What about all of the collectors?

Mpluya:
@MarkRx good question. So as far as the collectors that we're using go, the property that requires encryption is the database password.
Since the API's Docker file defines ENV SPRING_DATA_MONGODB_PASSWORD, that is what's been encrypted. This env variable then gets picked up by the other collectors through the docker link to the API via $HYGIEIA_API_ENV_SPRING_DATA_MONGODB_PASSWORD.

Let me know if you mean something else. Thanks!

api/docker/Dockerfile
```diff
@@ -25,4 +25,4 @@ VOLUME ["/hygieia/logs"]
 
 EXPOSE 8080
 CMD ./properties-builder.sh &&\
-  java -Djava.security.egd=file:/dev/./urandom -jar api.jar --spring.config.location=/hygieia/dashboard.properties
+  java -Djava.security.egd=file:/dev/./urandom -Djasypt.encryptor.password=$JASYPT_ENCRYPTOR_SECRET -jar api.jar --spring.config.location=/hygieia/dashboard.properties
```
jmcshane: For this `Dockerfile`, would you expected `$JASYPT_ENCRYPTOR_SECRET` to be provided in the environment on run?  If so, this should probably be added to the `api/README.md` file.

jmcshane:
I am very interested in this feature as well and would want it to be available for all Spring Boot apps within the project.  I think with what @Mpluya has provided (with the jasypt library as a core dependency), the `@EnableEncryptableProperties` just needs to be added to each collector.

The argument `jasypt.encryptor.password` can be provided in the `spring.config.location` property file, so it can be called out along with the other configuration parameters in each of the collector `README.md` files.  Then, there would not be a need to have updates to any of the Dockerfiles as it just sits with the other properties.

jmcshane:
@Mpluya I created a PR to your project in StateFarmIns to add these properties and a callout in each section of the README documentation, see StateFarmIns/Hygieia#4

api/docker/Dockerfile
Mpluya: @jmcshane yes the **JASYPT_ENCRYPTOR_SECRET** would be an environment property and this is reflected in the updates I made on api/README.md. Please go over the diff on the readme and let me know if there's anything lacking.

Mpluya:
@jmcshane you are correct.  Enabling property encryption on the collectors is just a matter of adding the @EnableEncryptableProperties annotation to the Application class on each collector. I did not add it by default to all the collectors, since I would like property encryption to be added on an as needed basis. I don't want an additional post processor kicking off on startup of each collector if there are no encrypted properties to deal with. I'm open to having it turned on on every collector if thats what we all agree to do. 

As for your pull request, I really appreciate you taking the time to enable property encryption on all collectors. I'd like for folks to weigh in on this before we put this in. 
As for putting the secret key (jasypt.encryptor.password) as another property, I would have to disagree with this approach. I would prefer for the secret key to be passed in either as a system property or environment variable when starting up your spring boot app. In my opinion, having an encrypted property and next to it is the secret key - pretty much gives away the the decrypted value. That said, in my organization when we startup Hygieia via docker containers, the secret key gets served up from a secure variable in our gitlab ci runner. 

I look forward to hear your thoughts on this. Thank you again.

jmcshane:
@Mpluya after thinking this over a little more, you are definitely right about not having the `jasypt.encryptor.password` property in the property file itself.

I think we just need to call this out explicitly in each of the `README.md` files and then, as you say, update the `Dockerfile` for each collector.

I'll make a modification to the PR I sent to remove the callout of the password in the properties and modify the above documentation section to talk about the system property.

tabladrum:
@satishc1 @nireeshT Lets line this up for testing.

Mpluya:
@satishc1 @nireeshT I just want to follow up how's testing coming along with this request. If there's any questions or anything I could clarify or help push this along, please let me know. Thanks.

satishc1:
LGTM

tabladrum:
@amitmawkin Please review.

tabladrum:
LGTM

PR #1005, opened 2016-12-04
Resolving capitalone/Hygieia#1004 

### LINKED ISSUE #1004: Deployment API endpoint does not create deployments that display properly
The deployment POST endpoint does not align with the way that the data is displayed in a number of ways:

1. The jobUrl property is completely ignored (search for references to  DeployDataCreateRequest::getJobUrl() in IDE).  There should be an option to access the job URL component by component.
2. The `appName`/`instanceUrl` property pair is being defined as distinct in the backend [here](https://github.com/capitalone/Hygieia/blob/master/api/src/main/java/com/capitalone/dashboard/service/DeployServiceImpl.java#L242) because the query looks to match the options passed in.  This means that two collectors can be created with the same name, which then appears as the same name in the Deploy widget dropdown.
3. The environmentUrl doesn't get set in the [createEnvComponent](https://github.com/capitalone/Hygieia/blob/master/api/src/main/java/com/capitalone/dashboard/service/DeployServiceImpl.java#L248) method.  This means that jobs submitted by the POST endpoint do not have a functional "View in Jenkins" link, as this is captured by the `environment.url` property [here](https://github.com/capitalone/Hygieia/blob/master/UI/src/components/widgets/deploy/detail.html#L25)

My suggested solution:
1. `jobUrl` property gets set on the environment_component collection object, then if this value is set it is used by the [deploy detail modal](https://github.com/capitalone/Hygieia/blob/master/UI/src/components/widgets/deploy/detail.html).  It will need to be at the component level, as the `jobUrl` property is specifying a deploy URL for a component.
2. Make `applicationName` distinct in the CollectorItem repo by removing `option.put("instanceUrl", request.getInstanceUrl());` from [DeployServiceImpl.java](https://github.com/capitalone/Hygieia/blob/master/api/src/main/java/com/capitalone/dashboard/service/DeployServiceImpl.java).  I realize this would be a breaking change for those with duplicate records here already by `applicationName` and `instanceUrl`.  I don't know what the project stance on this issue would be.
3. The `environmentUrl` should be set by the `instanceUrl` that is available in the API object.

I will be creating a PR for the related issues above.  I wanted the issue created to begin a discussion on these solutions.

jmcshane:
Resolving issues with deploy POST endpoint not creating Deploy records that display in a meaningful way.

1. Add jobUrl to the DeployableUnit
2. Provide a link to jobUrl on the component if present
3. Only search for new CollectorItem by appName, not also by instanceUrl
4. Use instanceUrl as environmentUrl

Resolves #1004


api/src/main/java/com/capitalone/dashboard/service/DeployServiceImpl.java
```diff
@@ -236,10 +222,7 @@ private CollectorItem createCollectorItem(Collector collector, DeployDataCreateR
         tempCi.setPushed(true);
         tempCi.setLastUpdated(System.currentTimeMillis());
         tempCi.setNiceName(request.getNiceName());
-        Map<String, Object> option = new HashMap<>();
-        option.put("applicationName", request.getAppName());
-        option.put("instanceUrl", request.getInstanceUrl());
-        tempCi.getOptions().putAll(option);
+        tempCi.getOptions().put("applicationName", request.getAppName());
```
MarkRx: CI's for difference instances are intentionally different. Removing the instanceURL will break this.

tabladrum: Agree with @MarkRx - this change will impact the uniqueness.

jmcshane: Ok, I'll put the line back in. The rest of the PR should still work, I'll review and submit changes this evening.

I think I was missing the niceName property option, that will work for my case.

jmcshane: See 592bcac

tabladrum:
PR 921 and this has conflicts. Need to resolve manually.

nireeshT:
@jmcshane - Add jobUrl to DeployableUnit - does this work for existing deployed units or only for the ones created new.

nireeshT:
@tabladrum @satishc1  - LGTM.

nireeshT:
screen shot 2017-01-31 at 1 42 57 pm
screen shot 2017-01-31 at 1 41 12 pm

Clicking on Dev1 navigated to details page with the component name linked to Jenkins Job url.

PR #1020, opened 2016-12-07
Configurable Pipeline Stage Names

MarkRx:
Added configurable pipeline stage names. Backwards compatible. Allows for more and different stage names. Also allows users to hide the Commit and Build stages.

Stage config defaults to:

> systemConfig.systemStages=${CONFIG_GLOBAL_SYSTEM_STAGES:-Commit,Build,DEV,QA,INT,PERF,PROD}

Assumptions:
- There is at most one "commit" stage and one "build" stage
- Users will not have more than 9 stages including build/commit (the UI column widths are not dynamic due to grids and I didn't see a need to make it dynamic)

feature1100-2

feature1100-3

feature1100-4_2


xpdable:
Sounds great. 
But where/which file I could change this ?

systemConfig.systemStages=${CONFIG_GLOBAL_SYSTEM_STAGES:-Commit,Build,DEV,QA,INT,PERF,PROD}

MarkRx:
The api properties file.

MarkRx:
rebased on to master 2.0.5

fluidjeel:
+1 , very much needed feature

tabladrum:
@MarkRx We are thinking that this should be configurable on the UI itself - per Dashboard. The solution should simply be 
1. Have the Env name an editable text box/label. 
2. Mapped environment can still remain drop down.
3. Save it as options of the widget itself. 

No api changes needed. All on the UI. 
Thoughts?

MarkRx:
That would not be enough. It wouldn't allow N number of stages (in our case we have 6 not 5). It would also require additional complexity in the product dashboard. Also, if we were to allow each team to have a different number of stages once again the product dashboard becomes a problem.

tabladrum:
Ok, I now see what you are saying. Let me play around it a little.

MarkRx:
I will investigate the per-team pipelines idea when I get time.

tabladrum:
@nireeshT has done some work... will share a pre-PR fork with you soon for testing it out.

MarkRx:
Will try tomorrow. Thanks!

MarkRx:
I see a number of issues:

- There are lingering references and logic referring to the System stages in the api settings. If any arbitrary stage can be created (as it seems is this case with the changes) then these need to be removed. If not then the UI needs to be tightened to make names of stages aliases to set system stages.
- There needs to be a way of having deterministic ordering on environment stages. This is needed in pipeline computations (such as finding the last stage a pipeline was in). If the intent is to remove the set global system stages then you will have to store ordering in the widget. The code as it is right now is not going to properly compute pipeline metrics (the mappings map in the widget has no intrinsic ordering). You should not relay on using a specific implementation (LinkedHashMap)
- My product dashboard is still computing based on the system stages set. If the intent is for any team to have any arbitrary stages then this needs to have some kind of computation to determine the stages to display

nireeshT:
@MarkRx  - I pushed latest changes to branch https://github.com/capitalone/Hygieia/tree/pr-1020 .

-> Removed instances of api systemConfig.
-> Used the ordering for stages in sequence they are added. Used the same order to display on Product dashboard.
Check it out and let me know if you see issues.

fluidjeel:
@MarkRx @nireeshT @tabladrum Bumping this thread. Badly needed feature for us. Can you please confirm is there is any plan to test the merged changes and make it part of main release.

nireeshT:
@fluidjeel  - this is merged to master. Please update your code base.

PR #1021, opened 2016-12-07
Fix for product dashboard sometimes not displaying data right away #864

MarkRx:
Fixes the dashboard sometimes not showing data right away due to a timing issue. I think it has to do with the order the asynchronous rest calls come back in. This wraps the setTeamData method into a $timeout function which puts it at the end of the event queue and causes a $scope.$apply() to be run after it is finished. I don't know if the $timeout wrap is necessary (as opposed to using $scope.$apply() but that is what other js functions do so I followed suit.

### DIFF
--- a/UI/src/components/widgets/product/js/commit-data.js
+++ b/UI/src/components/widgets/product/js/commit-data.js
@@ -17,6 +17,7 @@
         var db = dependencies.db,
             configuredTeam = dependencies.configuredTeam,
             $q = dependencies.$q,
+            $timeout = dependencies.$timeout,
             isReload = dependencies.isReload,
             pipelineData = dependencies.pipelineData,
             nowTimestamp = dependencies.nowTimestamp,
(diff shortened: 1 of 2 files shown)


MarkRx:
rebased on to master 2.0.5

supremex1:
@MarkRx  The product dashboard is still not displaying the details after opening & closing the configuration multiple times. Can you suggest what should I do in the dashboard to refresh or to display the details?

supremex1:
@MarkRx Whether we have to manually change these 2 files or have you committed the edits? Kindly provide me the link to download the latest one if commit has been done. Thanks.

MarkRx:
Be sure that there actually is data to display. Right now pipelines will only start gathering data after they are created.

supremex1:
@MarkRx Yes, I have created the pipeline too and it is successfully displaying in team dashboards but same DEV, INT .... data is not displaying in product db.

MarkRx:
@supremex1 If data does not display after 2 minutes this means your problem is with your pipeline and is unrelated to what this PR is attempting to fix.

vetrivelan25:
Hi @MarkRx ,
                     That Merge was completed on master branch reg Product dashboard doesn't display correct data issue.If its merger please send the latest link.

MarkRx:
Fixed in master.

Issue #1028, opened 2016-12-12
Jira Colloctor is retrieving nothing without Team plugin?


xpdable:
See my comments at the end of #974 

After I moved to the latest version of Jira-Collector with a seems-to-be mandatory Team plugin, all new-added data in my jira is not retrieved into mongodb even I set some fake field in config file for the team.
My Jira does not have that plugin and no plan to install it.
However we'd like to use the new function of the stastics in Issue.
Could you help on this?

MarkRx:
@pxk5958

MarkRx:
PR #1030

tabladrum:
We are encountering some issues while testing out this PR. More details to come @satishc1

tabladrum:
We are encountering some issues while testing out this PR 1030. More details to come @satishc1

satishc1:
@xpdable Functionality merged in https://github.com/capitalone/Hygieia/pull/1137 Use latest codebase and follow cleanup instructions provided in PR.

xpdable:
@satishc1 @tabladrum @MarkRx  I've moved to latest codebase, Jira collector seems works with some exception.

> 2017-02-05 14:08:00,002 INFO  c.c.d.collector.CollectorTask - Running Collector: Jira
2017-02-05 14:08:00,004 INFO  c.c.d.collector.CollectorTask - -----------------------------------
2017-02-05 14:08:00,004 INFO  c.c.d.collector.CollectorTask - http://jirag.int.thomsonreuters.com/
2017-02-05 14:08:00,005 INFO  c.c.d.collector.CollectorTask - -----------------------------------
2017-02-05 14:08:00,005 DEBUG c.c.d.client.team.TeamDataClientImpl - Constructing data collection for the feature widget, team-level data...
2017-02-05 14:08:00,551 ERROR c.c.d.client.DefaultJiraClient - IOException
java.io.FileNotFoundException: http://jirag.int.thomsonreuters.com/rest/tempo-teams/1/team
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(Unknown Source) ~[na:1.8.0_91]
	at sun.net.www.protocol.http.HttpURLConnection.getInputStream(Unknown Source) ~[na:1.8.0_91]
	at java.net.URLConnection.getContent(Unknown Source) ~[na:1.8.0_91]
[...]

Issue #826, opened 2016-09-15
Bug: Saving a Code Repo widget after data has been collected results in duplicate collector_items


MarkRx:
Duplicate collector_items can be created which results in strange behavior. Below is an example for SCM collector_items but I suspect this may happen in other places.

Steps to reproduce
1. Create a dashboard and set it up the "Code Repo" widget to use a repository that has not been used.
2. Wait for a scm collector to run
3. Refresh the page. Open the widget and save it.

There will now be 2 collector_items that correspond to the same repo + branch + scm type combination.

This is because once the scm collector runs the collector might have additional options tacked on. For example, the BitbucketCollector adds the "lastUpdate" and "lastCommit" options. Hence findByCollectorAndOptions does not find the existing collector_item since the incoming item.getOptions() do not contain this information.
CollectorServiceImpl.java:

```
    public CollectorItem createCollectorItem(CollectorItem item) {
        CollectorItem existing = collectorItemRepository.findByCollectorAndOptions(
                item.getCollectorId(), item.getOptions());
        if (existing != null) {
            item.setId(existing.getId());
        }
        return collectorItemRepository.save(item);
    }
```

I think findByCollectorAndOptions needs to be changed to allow items to show up if they have additional properties that do not exist in the incoming map. Will have to check mongo documentation if this can be done.

MarkRx:
A sparse unique index (or partial if minimum mongo version is 3.2) could be added to collectoritem for options:
{'options.scm':1,
'options.branch':1,
'options.url':1}

I think findByCollectorAndOptions would have to be manually implemented to search for the properties in the map.

MarkRx:
Looks like when changing settings we are calling CollectorServiceImpl /collector/item which is always creating a new collector item overwriting an old one if it already exists. As a result the additional metadata information is lost (such as last update time).

Sbrenthughes:
@MarkRx can you confirm this issue has been resolved in the latest release?

tabladrum:
I don't see this occurring any more.

PR #1030, opened 2016-12-12
Fix for jira collector exception if it does not have team-tempo plugin #1028

MarkRx:
#1028 

@pxk5958


MarkRx:
Added null guard for when jira instance is < 6

MarkRx:
rebased on to master 2.0.5

MarkRx:
Please merge this. It adds NPE checks to fix some exceptions when people do not have the team tempo plugin installed.

collectors/feature/jira/src/main/java/com/capitalone/dashboard/client/team/TeamDataClientImpl.java
```diff
@@ -63,13 +60,6 @@ public int updateTeamInformation() {
 		if (CollectionUtils.isNotEmpty(teams)) {
 			updateMongoInfo(teams);
 			count += teams.size();
-		} else {
-			List<BasicProject> projects = jiraClient.getProjects();
-			
-			if (CollectionUtils.isNotEmpty(projects)) {
-				updateMongoInfoLegacy(projects);
-				count += projects.size();
-			}
```
tabladrum: Why did we take this code out? If the tempo team plugin is not there, it wont create collector item anymore with this code gone.

MarkRx: The original idea was if the team tempo plugin is not there then the database should not collect any team data. Instead the user will select the "ANY" field in the team dropdown. Right now with that code block a half baked attempt was made for backwards compatibility to store project data in both the project and the team repos. We figured we should remove that because it was confusing.

In reviewing and testing this code after being removed from it I realize now that that may be flawed as not having a team (ScopeOwnerCollectorItem) results in the UI not selecting a collector item. The "legacy" approach above may also be flawed as it may result in two dropdowns for "team" and "project" that are the same (both "project") and I don't remember if we had put in something for that case.

I will look more into this tomorrow.

tabladrum: That's exactly the issue that we have hit. We have another issue - not related to this change. For Kanban in Jira there may not be any estimate what so ever. We are thinking if that is the case, we should display simply the number of issues instead of points or hours. Thoughts?

MarkRx: PR #1054 does just that

tabladrum: Ah. :-)

MarkRx: @tabladrum what do you think of the idea of having a pseudo scopeowner (Collector item) representing the "any" option? This way we could still have an attached collector item even if "Any" is selected (if there are teams and any is selected or if there are no teams and it is the only option). It works because the feature documents are not attached to collector items and instead we query them by fields.

tabladrum: Thinking out loud - there has to be a project or a team. Why not store either or both in teamID and projectID (a new field)? One of these must have a selection value.

MarkRx: it does. When there is no team though ("Any") there is no collector item saved. Projects don't use collector items they use the scope repo.

MarkRx: I think the current approach (leaving it how it is in this PR) would work if I fix the feature service to find the Jira collector record directly instead of attempting to go through the attached collector item.

satishc1: @MarkRx In the case where someone would have tempo teams in Jira, do they need to select Project in addition to Team in the Hygieia feature config dropdown ? We were trying to decide whether to display the Project field if Tempo teams exist in Jira.

MarkRx: They do. In jira teams and projects are independent. A team can belong to multiple projects. A project can have multiple teams. Hence both inputs are needed as these are all valid filters:
- Project: Any, Team: myteam
- Project: myproject, Team: Any
- Project: myproject, Team: myteam
- Project: Any, Team: Any

tabladrum: perfect.

MarkRx: I'm trying something now. I will store the collectorId in options when the team selection is "Any". It will then be passed into the feature rest calls which will have two service variants: get*ByComponentId and get*ByCollectorId. The get*ByComponentId will simply extract the collectorId from the component and call the corresponding get*ByComponentId

MarkRx:
Pushed changes. Added information about the collectorId when there is no collectorItem (when team is 'Any'). UI will use the collectorIdOverride when necessary.

There is one outstanding issue - the config does not show up with the previously saved options if "Any" for team is selected. I have not looked in to why yet.

MarkRx:
Ok found out why that was happening.

Code should be good now. Go ahead and give it a try. To emulate not having teams simply remove them from the database or comment out the part of the team collector that runs.

collectors/feature/jira/src/main/java/com/capitalone/dashboard/client/team/TeamDataClientImpl.java
satishc1: @MarkRx @tabladrum 
While testing the latest changes, I was comparing the jira and versionone collectors and noted some differences in the way the data is being collected and shown in the UI.
- No collector items get created when the jira collector is run as opposed to versionone.
- The components collection that is created when you save the jira config doesn't have the corresponding collector_item
Components
/* 1 */
{
    "_id" : ObjectId("588187ebf027536d790834a4"),
    "_class" : "com.capitalone.dashboard.model.Component",
    "name" : "DashAppName",
    "collectorItems" : {}
}
- The widget inside the Dashboard collection is being tied directly to the Collector via "collectorIdOverride" : "58818744f027536d1b592e96". In VersionOne we tie the widget to the component which in turn ties to the Collector.

Should we be consistent in the jira / versionone designs ? It will also be easier to refactor in the future.

MarkRx: First bullet - Yup. I avoided changing the versionone collector because I do not have the ability to test changes to it. However, doesn't versionone always have teams? In that case your first difference is legitimate. 

The last two bullet points happen for both versionone and jira. There is no specific logic in the UI to differentiate them. Since teams are stored in collector items there is no team to associate the component with when 'Any' is selected. The override is used as a way to get back to the collector. It is used in the feature API calls which use the feature repository which are all implementation agnostic.

tabladrum: This is what I am thinking in terms of making it uniform across VersionOne and Jira and also matching the overall design pattern of rest of Hygieia.

1. Do not create any collector item from either VersionOne or Jira collector. Only create team and/or scope documents. 
2. UI drop downs fetches data from scope (for Project drop down) and team (for team drop down)
3. Once UI config is saved, create a ScopeOwnerCollectorItem with the combination configured. The collector item will have the project and team selected on the UI. One of these has to be present, both can't be null (error validation on UI)
4. Jira collector and VersionOne collector collects feature data based on the project+team in the collector item collection. This can potentially save a lot of time in data collection too.

satishc1: @MarkRx @tabladrum If we agree on this proposed design, I can work on implementing it. I do agree there are parts of 1030 we should incorporate into it.

MarkRx: For 1 - So would you take teams out of collector_item and put them into their own repo?
For 2 - I don't think issue data should be separated out by team / project since there would be a lot of duplication. There should be a team repo (listing all teams), a project repo (listing all projects), and a feature repo (having all issues). Features then have IDs that relate them to their team / project / neither
For 3 - I don't agree. Both can be 'Any' to display everything. These are simply filters into the feature repo. There could be more filters in the future so we wouldn't want to segregate data but instead keep them in feature
For 4 - Sounds fair but may be a lot of work. I assume all features will still be shared?

tabladrum: What is the use case of Any Project - Any Team combination? Basically saying show me everything in Jira?

MarkRx: I may have misread what you said. Is this what you are suggesting?

Move teams out of collector_item and into a new repository (handled similar to the scope repo). Continue to put all features in the feature repo but only load them on-demand based on UI selections as opposed to gathering everything at once. Have the UI create a collector item representing the config in a similar manor to how the git collector is handled.

MarkRx: Yes 'Any' 'Any' would show everything. I don't know if it is a valid use case or not. The ability exists right now. It wouldn't bother me to remove that combination if it can let us save a lot of time collecting issues.

tabladrum: Yes. Exactly. Have a new collection called "team" - just like "scope". No change in feature repo. Create collector item just like github config. API will pull data from feature based on the collector item's project/team value and apply view options on top of the data gathered.

MarkRx: Is it correct to say that the feature documents will NOT link back to a collector item? I wouldn't want  (projectA, teamX), (projectA, Any), (Any, teamX) to result in a bunch of duplicated features. The git collector suffers from this problem - there are a lot of duplicate commits because they are separated out by collector_item. In our case since we query on-demand based on parameters instead of by ids we can share the data.

tabladrum: Feature documents point to collector, but not collector item. 
With that, we will not have duplication. SCM commits are little tricky - specially git. The same commit "travel" though branches. I have not seen duplicates in terms of "repo/branch/sha" combination. If there is we need to correct that.

MarkRx: Ok thanks for the clarification. Yea that sounds good.

repo/branch/sha duplicates occur due to #826. @pxk5958 had researched it.

tabladrum: Yes. The collector items are duplicated. Need to fix that. The commits are unique.

tabladrum:
@satishc1 is working on a full refactor on this based on the discussions in this PR. FYI.

collectors/feature/jira/src/main/java/com/capitalone/dashboard/client/team/TeamDataClientImpl.java
satishc1: @MarkRx @tabladrum I've made the necessary changes to refactor the Jira and VersionOne collectors. The change to collect features on-demand will be made after we merge the first set of changes.

Can you take a look at my fork and pull the code down and try it out before I send in the PR. The code that I've pushed to my fork includes parts of PR-1030 coupled with the re-factoring, so essentially it will replace PR-1030.  Once we review and merge the code from my fork, we will need to change PR-1054 as needed and merge it.

Steps to re-run the jira / versionone collectors:
1. Make a backup of your database.
2. delete collector_items where collectorId in (select collectors.id from collectors where collectors.name in ("Jira", "VersionOne") )
3. Drop feature , scope , team
4. Run jira and versionone collector
5. Open your dashboard and re-configure your feature widget to pick the right project / team value.

Script for Steps 2 and 3 is given below.

```
db.collectors.find({
	"name" : { $in : ["Jira", "VersionOne"] }   
}).forEach(
	function(myDoc) {
		db.collector_items.remove( {"collectorId" : myDoc._id} );
	}
);

print('Dropping feature collection');
db.feature.drop();

print('Dropping scope collection');
db.scope.drop();

print('Dropping team collection');
db.team.drop();
```

MarkRx: Are you intentionally repurposing ScopeOwner to correspond to the new team repo? I think you need a new CollectorType for the collector_item representing the filtering options chosen. Right now the Feature service is pulling the ScopeOwner collector item off of the component document. Instead it should be pulling something else.

satishc1: @MarkRx @tabladrum Thanks. Removed the confusing Feature / ScopeOwner references and replaced with AgileTool. Added some more cleanup and committed. Please take a look.

satishc1:
@MarkRx @tabladrum This PR will be closed and replaced with https://github.com/capitalone/Hygieia/pull/1137

PR #1036, opened 2016-12-14
changes to allow collecting builds from a workflow type jenkins job

Mpluya:
Problem: When the Jenkins collector runs and encounters a workflow type job ("_class":"org.jenkinsci.plugins.workflow.job.WorkflowRun"), the collector throws an exception since workflow type jobs don't have changeSets. This is a fix to add a null check and immediately exit from the method when there are no changeSets to build, but still allow to collect build information from a workflow job.

### DIFF
--- a/collectors/build/jenkins/src/main/java/com/capitalone/dashboard/collector/DefaultHudsonClient.java
+++ b/collectors/build/jenkins/src/main/java/com/capitalone/dashboard/collector/DefaultHudsonClient.java
@@ -1,11 +1,23 @@
 package com.capitalone.dashboard.collector;
 
-import com.capitalone.dashboard.model.Build;
-import com.capitalone.dashboard.model.BuildStatus;
-import com.capitalone.dashboard.model.HudsonJob;
-import com.capitalone.dashboard.model.RepoBranch;
-import com.capitalone.dashboard.model.SCM;
-import com.capitalone.dashboard.util.Supplier;
+import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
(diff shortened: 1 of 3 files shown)


amitmawkin:
@Mpluya Does this take care for Jenkins 2.0 pipelines too, or just the workflow plugin.

collectors/build/jenkins/src/main/java/com/capitalone/dashboard/collector/DefaultHudsonClient.java
```diff
@@ -232,6 +233,7 @@ public static String rebuildJobUrl(String build, String server) throws URISyntax
      */
     private void addChangeSets(Build build, JSONObject buildJson) {
         JSONObject changeSet = (JSONObject) buildJson.get("changeSet");
+        if(changeSet == null) { return; }
```
MarkRx: This only prevents a NPE. It does not correctly gather information

MarkRx:
@pxk5958 has a PR coming soon that will support the pipeline plugin

amitmawkin:
Thanks @MarkRx

MarkRx:
#1038

Mpluya:
@amitmawkin this solves the error encountered on any workflow type job with just an addition of a null check. I probably should reword this request to instead say "handle null changeSet scenarios",
 which i happen to have encountered on workflow type jobs. Thanks.

pxk5958:
@Mpluya Workflow/Pipeline jobs have the changesets as a JSONArray of "changeSets" instead of "changeSet" in normal jobs. That's why you're encountering the NPE. With your change, it would skip the NPE, but will not collect the actual changesets if present.

Mpluya:
@pxk5958 I appreciate the explanation. Closing this request since PR #1038 addresses the entire need.

PR #1042, opened 2016-12-19
Updated the pom to 2.0.5-SNAPSHOT and cut a release

amitmawkin:
https://github.com/capitalone/Hygieia/releases/tag/Hygieia-2.0.4

### DIFF
--- a/UI/pom.xml
+++ b/UI/pom.xml
@@ -9,7 +9,7 @@
   <parent>
     <groupId>com.capitalone.dashboard</groupId>
     <artifactId>Hygieia</artifactId>
-    <version>2.0.4-SNAPSHOT</version>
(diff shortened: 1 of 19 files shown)


MarkRx:
Does the jenkins plugin version need to correspond to the Hygieia version or is that on its own lifecycle?

amitmawkin:
@MarkRx we made a decision a while back to make all version of each component same for ease of support . 
@tabladrum Do you think we should separate the Jenkins plugin and manage it separately

tabladrum:
At this point, we are ready to make them all even and make Jenkins plugin part of the main project and publish to maven central.

satishc1:
LGTM

tabladrum:
LGTM

PR #1045, opened 2016-12-20
Disabled browser-sync ghost mode, by default

### LINKED ISSUE #1016: Is BrowserSync ghostMode necessary?
When we serve the UI with gulp serve (without nginx), the interaction synchronization through [ghostMode](https://www.browsersync.io/docs/options#option-ghostMode) option causes problems. It is not working as intended and browsers end up in different pages sometimes and cause problems during testing. For example, we consistently end up in a scenario where one browser goes into a dashboard page whereas another browser remains in the home page (site.html), and eventually interactions trigger the deleteDashboard() javascript function on the latter browser. We consistently get a dashboard deleted when the Monitor widget's configure wheel button is clicked in the first browser. That is not the only case where dashboards get deleted though.

So the question is, was ghostMode intended to be used? If not, we could turn it off in gulpfile.js when initializing browser-sync, since it's causing the dashboards to be deleted during testing, which is rather annoying.

Other issues reporting similar problems: [#871](https://github.com/capitalone/Hygieia/issues/871), [#627](https://github.com/capitalone/Hygieia/issues/627)

@mrpudn @amitmawkin @tabladrum

mrpudn:
Since this has been the source of several issues, I think it best to disable browser-sync's [`ghostMode`](https://www.browsersync.io/docs/options/#option-ghostMode) feature, by default.

Ghost mode can be enabled by running `gulp serve:ghost-mode` if you require this functionality.

Resolves #1016

### DIFF
--- a/UI/README.md
+++ b/UI/README.md
@@ -62,6 +62,11 @@ Local Testing with Mocks:
  gulp serve --local true
 ```
 
+Using browser-sync's [`ghostMode`](https://www.browsersync.io/docs/options#option-ghostMode) functionality:
+```bash
+gulp serve:ghost-mode
+```
+
 or you can run via maven from UI project root folder
  ```bash
  mvn clean package integration-test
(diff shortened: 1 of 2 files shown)


satishc1:
LGTM

tabladrum:
LGTM

tabladrum:
@amitmawkin please merge.

PR #1027, opened 2016-12-09
Angular/UI-Bootstrap Update to latest

rob-miller-777:
Updated Angular version to latest stable release of Angular (1.5.9), as well as updated version of UI-Bootstrap (2.3.0).  

Most of the changes that had to be made outside of the version changes revolved around Angular Bootstrap, and Angular Cookies. The bootstrap dependency started prefixing their modules with "uib", which accounts for a significant portion of the changes. Angular changed their cookie api to use "getters" and "puts" instead of accessing the variables directly.

PR #1049, opened 2016-12-22
Pagination for /api/collector/item/type/*

brennanananan:
When having a large amount of collector items for a specific collector type (10000 Jenkins Jobs for instance), it could take some extra time to send all of the items to the UI. This pull request allows the api to return paginated results. We also repurposed the 'collectorItemsByType' operation to allow for a search/filter functionality. This is backwards compatible with all widgets and is a non breaking change; the only changes to the response from the API are in the header dictating paging information. If no paging information is included in the request, all items are returned by default.

We will have PR's in the future that allow these widgets to have a 'typeahead' functionality that will dynamically load the field as you type, which utilize the changes made in this PR. This will all result in a significant performance increase and enhance user experience in the case of large collector item sets.


MarkRx:
@pxk5958 is actually investigating this right now. We also have 10000's of Jenkins jobs and the build widget config in-particular is not very responsive. We have found two causes:

- The angular module ui-select has several well known performance problems. @pxk5958 was looking into using an alternative called selectize.
- For some reason spring is taking a long time to serialize the database results into objects. Results return quickly when using a mongo utility.

Does paging really help? I don't think the actual network load is an issue. You might be seeing a perceived increase in performance because the reduction in data addresses reduces the effects of both issues above.

brennanananan:
This PR alone will have almost zero effect on performance, by we will have PR's coming early next year that will utilize this - greatly improving user experience when the job list is quite large. If you prefer, we could bundle the two PRs together. We split them up initially between server side changes and UI changes. Note that our UI changes depend on #1027  @rob-miller-777 @Mpluya

rob-miller-777:
Like @brennanananan said, we have another merge request that will take advantage of this one.  We changed the ui-select to Angular-Bootstrap typeahead.  We filter the results based on input from the user, then paginate those results to only return the top 20.  Load time are almost instant now.

amitmawkin:
nice guys, ui-select has issues with large datasets, so waiting for the PR to test it out.

mdskd:
@amitmawkin would you like us to submit the other PR before accepting this one?

mdskd:
Or combine the two into a single pull request? Sorry, can't edit comment on mobile.

amitmawkin:
@mdskd  please keep it seperate, this pR is getting further tested by @satishc1

nireeshT:
@tabladrum  - PR  reviewed and tested - OK.

mdskd:
any feedback on this PR?

thanks!

mdskd:
@tabladrum @amitmawkin is there anything we can help with on this issue? or outside delays holding up accepting this PR?

thanks!

tabladrum:
We are testing and refactoring two big PRs now.. as soon as we are done, we will merge this before merging those. Just wanted to make sure things don't go into conflict.

tabladrum:
@brennanananan can you rebase?

rob-miller-777:
Done

satishc1:
@tabladrum LGTM

Issue #940, opened 2016-11-10
Jira - Track Number of Tickets


leahxu:
Enhancement request, thanks!

Jira gives you the ability to turn off ticket estimate, it should default to tracking the number of issues in each state in that case.

xpdable:
This's absolutely what we'd love to have too.
Something like a list with a jira filter result with <id>+<summary>+<status>
For example, the similar one with Jira dashboard:

![image](https://cloud.githubusercontent.com/assets/2903719/20874828/8a1e48c6-baf0-11e6-9713-f26d249fd17e.png)

xpdable:
Sorry for chasing anywhere, but do we have any plan on this feature?

MarkRx:
You can just add another option - hours, storypoints, and count. Do what we did for hours/story points and tweak the api

xpdable:
@MarkRx I done the issue count on my side.
![image](https://cloud.githubusercontent.com/assets/2903719/21445191/dc171136-c8f0-11e6-9c15-9841ba132459.png)

What I did is simply trick getSprintEstimates() to return the issue count in the SprintEstimate object where I get the issue count, as what hours/sp do, inside the switch case statements in file FeatureServiceImpl.java

As we seldom have the usage for Hours/sp in jira, I just did kind of hard-code to return only issue count. 
However, I am still look forward to know your plan of the official implementation on this or even I could help ?

MarkRx:
PR #1054

MarkRx:
This feature has been merged into master with PR #1054.

satishc1:
@leahxu Functionality merged in https://github.com/capitalone/Hygieia/pull/1054 Use latest codebase. You’ll also need to follow cleanup instructions provided in https://github.com/capitalone/Hygieia/pull/1137

PR #1054, opened 2016-12-27
Feature widget - add ability to track number of issues instead of estimates - #940

MarkRx:
The feature widget now has an option to display metrics based on the number of tickets instead of estimates. This is for setups that do not use estimates.

@pxk5958 

#940


Mahendra1260:
@MarkRx 
is this is merged with Master branch? Can it be used with latest code?

MarkRx:
It is not in master. You will have to bug @tabladrum or @satishc1 lol

satishc1:
LGTM

Issue #1048, opened 2016-12-21
Jenkins Plugin: Jenkins 2.0 support


tabladrum:
Here is my fork where I am pushing code to have pipeline support to Jenkins plugin. I have so far done build and artifact publishing. The changes should be backward compatible. Will appreciate if anyone has time to do some testing and pre-review.

https://github.com/tabladrum/Hygieia/tree/master/hygieia-jenkins-plugin

@MarkRx Question for you: the Metadata section in ArtifactBuilder will not work for pipeline. Instead of getting the SCM metadata from envVar, I will get it from the build data that I already have in that section of the code. May be a second thought on the meta data? Do we need it? Or should reduce duplication and have only the meta data? In any case, pipeline api returns a list of SCMs and I will have to handle it as a list - which means the SCM metadata must be a list. Thoughts?

MarkRx:
How does the jenkins plugin handle multiple scms right now? Are we just using the first?

The metadata is needed. It wouldn't make sense for metadata to be a list as it is simply a key-value property holder (that corresponds to artifactory properties). Instead you could use a numbering scheme similar to how jenkins handles multiple repositories. Example:

```
scmUrl: "url",
scmUrl1: "url1",
scmUrl2: "url2"
scmBranch: "..."
scmBranch1: "..."
```

@pxk5958 did you encounter this with #1038?

tabladrum:
Currently it stores the SCMs in a list of CodeRepo structure. For most cases it is just one. Not sure if any one builds multiple repos at once.

tabladrum:
https://github.com/capitalone/Hygieia/pull/1055

tabladrum:
Closing this as the PR is in.

PR #1055, opened 2016-12-27
Jenkins 2.0 support in Hygieia Jenkins plugin

tabladrum:
This is a working version. Wanted to open up for review and suggestions. There are scopes of refactoring (removing code duplications) but will get to that later.


tabladrum:
@amitmawkin @satishc1 - I have tested this with 2.35 version of Jenkins 2.0 and multiple versions of Sonar (for Sonar publish). Can you test with your version of Jenkins?

hygieia-jenkins-plugin/README.md
```diff
@@ -2,8 +2,7 @@
 
 # Developer instructions
 
-Install Maven and JDK.  This was last built with Maven 3.2.5 and OpenJDK
-1.7.0\_75 
+Install Maven and JDK.  This was last built with Maven 3.2.5 and JDK 1.8.
```
amitmawkin: @tabladrum can you see if you can build this with maven 3.3.9 . The rest of the project was built with that version of maven

hygieia-jenkins-plugin/src/main/java/hygieia/builder/ArtifactBuilder.java
```diff
@@ -66,30 +89,44 @@ private void buildArtifacts() {
...
+                bac.getMetadata().put("buildUrl", HygieiaUtils.getBuildUrl(run));
+                bac.getMetadata().put("buildNumber", HygieiaUtils.getBuildNumber(run));
+                bac.getMetadata().put("jobUrl", HygieiaUtils.getJobUrl(run));
+                bac.getMetadata().put("jobName", HygieiaUtils.getJobName(run));
+                bac.getMetadata().put("instanceUrl", HygieiaUtils.getInstanceUrl(run, listener));
+
+
+                if (run instanceof AbstractBuild) {
+                    AbstractBuild abstractBuild = (AbstractBuild) run;
+                    String scmUrl = HygieiaUtils.getScmUrl(abstractBuild, listener);
```
MarkRx: does WorkflowRun not have this information?

tabladrum: Workflow does not have those environment variables. I will have to fill those in in a different way. But, the question is this: How do we handle multiple SCM repo build? Thoughts?

tabladrum: Workflow does not have those environment variables. I will have to fill those in in a different way. But, the question is this: How do we handle multiple SCM repo build? Thoughts?

MarkRx: Right now we don't with the normal jenkins job so you can just use the first one listed. If you feel adventurous though just add a number to the end of each one after the first one : https://github.com/capitalone/Hygieia/issues/1048#issuecomment-268658328

PR #1068, opened 2017-01-04
Client pagination

Mpluya:
This harvests the enhancements in PR https://github.com/capitalone/Hygieia/pull/1049 and switches the config UI on both Jenkins and Sonar (widgets that deal with a massive data set) from ui-select to  an input with uib-typeahead, returning 20 results at a time. In order to use uib-typeahead, upgrading to the latest stable release of Angular (1.5.9) and latest version of UI-Bootstrap (2.3.0) were necessary.


mdskd:
@Mpluya can you resolve conflicts please?!

@tabladrum @amitmawkin any feedback on client pagination?

Mpluya:
@mdskd conflicts resolved.

satishc1:
@Mpluya Thanks for the contribution. Can you please add a screenshot and some details on how the UI looks with this change? If you search for something in the ui select dropdown, and if it returns the first 20 hits, is there a next (prev) button for the user to select and look through the results ?

Mpluya:
@satishc1 below is a screenshot of how configuring jenkins or sonar looks like with this enhancement. 

The way we (@rob-miller-777 and I) have it configured is:

- on click to the search field without typing anything, it will send a server side request without any search criteria but the size of 20 
`api/collector/item/type/build?size=20`
- with the screenshot I provided below I typed in 'BusinessLine' as a search criteria. As I was typing this, the following server side requests were made:
`api/collector/item/type/build?search=Busines&size=20`
`api/collector/item/type/build?search=Business&size=20`
`api/collector/item/type/build?search=BusinessLine&size=20`
  This is so, because we've configured the search text field with a 250 ms of typeahead-wait-ms. Typeahead-wait-ms (from angular docs) is the minimal wait time after last character typed before typeahead kicks-in.

The idea is for users to type in what they are searching for and if the result they are looking for doesn't show up on the first 20 results, continue to refine the results by typing more.

Let me know if you need further details. Thanks.

buildwidget_configure

satishc1:
@Mpluya @rob-miller-777 @mdskd @tabladrum Thanks for the screenshot and the details. If the results are being paginated, there should be a prev and next button to browse the "pages". The user will not know why their "desired" result is not showing up in the first 20 or whatever number we set. While this is a great feature we really need it, can you explore the possibility of adding a way to browse through the search results ? Your thoughts ?

rob-miller-777:
Yeah, I think it could benefit from some sort of infinite scroll.  I think we had looked into that, but finding a library that could handle the filtering and infinite scroll was challenging.  We could take a look again however.

mdskd:
agree with comments that some sort of paging in the ui would be good - infinite scroll fits more in this particular scenario from a ui design perspective.


Let take a look at this next week @rob-miller-777

rob-miller-777:
@satishc1 @tabladrum We have been trying a few different things to utilize pagination.  Our first attempt was to add infinite scroll to the UIB-Typeahead directive.  This doesn't seem to be possible without either modifying the source code for the typeahead directive, or creating our own.  Neither of which seem ideal to us.  We also couldn't find any existing libraries that combine infinite scroll/pagination and filtering in a select dialog.  We are currently trying a different approach by adding a table/list of "jobs" to the form, and server side pagination and filtering on top of that.  The functionality is much easier to implement for this, however the user experience suffers a little bit as it takes up more of the screen, and there are more buttons to click for navigating.  So the two options we are looking at right now are keeping the functionality as is in this merge request, where to find your job, you have to continue to type your job name until it appears in the list.  Or we can add the table with filtering and pagination.  We will submit a merge request with the second option when ready.

satishc1:
@rob-miller-777 @tabladrum  If the UIB-typeahead paginated scrolling is tough to implement, then given a choice between the 2 options, I would pick option 1. We can have a quick conference call if you have some time available. We are on gitter.im, if you wanna ping me at a convenient time. Thanks for researching this.

rob-miller-777:
blurredss

Here is what option 2 could look like.  Could make some small tweaks, but something along these lines.  I don't think I'll be able to use gitter.im right now since it hasn't been approved here for business use.  So for the time being, would you mind just communicating here?

rob-miller-777:
Just to be clear on the two options. 

1. Functions as is in this current merge request.  No pagination.  Have to type in the name of your job until it appears.

2. List of jobs which can be filtered and paginated through an input box and pagination directive. 

Pictures of both above.

satishc1:
@rob-miller-777 Looking at the screenshot for Option 2, I think Option 1 is definitely easier to use. Let me pull the PR down and play around with it a little more. At the moment i'm tending towards merging the PR as is. Let me discuss it internally here and i'll let u know soon. Thanks again for your team's effort on this. We truly appreciate the collaboration.

rob-miller-777:
Sounds good, appreciate the feedback.  If you want to play around with option 2, let me know and I can get a branch out there.  Still needs some clean up, but functions properly.

UI/src/components/widgets/deploy/config.js
```diff
@@ -5,9 +5,9 @@
         .module(HygieiaConfig.module)
         .controller('deployConfigController', deployConfigController);
 
-    deployConfigController.$inject = ['modalData', 'collectorData', 'systemConfigData', '$modalInstance', '$q', '$scope'];
 
-    function deployConfigController(modalData, collectorData, systemConfigData, $modalInstance, $q, $scope) {
+    deployConfigController.$inject = ['modalData', 'collectorData', 'systemConfigData', '$uibModalInstance', '$q'];
```
satishc1: @Mpluya Will taking out $scope affect the $scope.reload down below?

UI/src/components/widgets/deploy/config.js
```diff
@@ -5,9 +5,9 @@
...
+    function deployConfigController(modalData, collectorData, systemConfigData, $uibModalInstance, $q) {
```
satishc1: @Mpluya Will taking out $scope affect the $scope.reload down below?

satishc1:
@Mpluya Need to add the $scope in the deploy/config.js in the inject and the function.
ReferenceError: $scope is not defined
    at new deployConfigController (config.js:143)
    at Object.invoke (bower.js:6178)
    at $controllerInit (bower.js:11999)
    at resolveSuccess (bower.js:49359)
    at processQueue (bower.js:18103)
    at bower.js:18119
    at Scope.$eval (bower.js:19401)
    at Scope.$digest (bower.js:19215)
    at Scope.$apply (bower.js:19509)
    at done (bower.js:13489)

Mpluya:
@satishc1 thank you.

satishc1:
@tabladrum LGTM

Issue #1067, opened 2017-01-04
Question: Dashboard name validation


Mpluya:
We ran into a scenario where one of our users created a dashboard with a slash ('/') in the name, i.e. "Test/Dashboard". While dashboard creation was successful, he then was unable to configure any of the widgets. This is so because the configure action was unable to hit the '/dashboard/myowner/{dashboardtitle}' endpoint, since his request was '/dashboard/myowner/Test/Dashboard'. My question to the group is, would there be interest in adding a validation around the dashboard name to prevent slashes and or any special characters? If yes, I'll open a PR on this. Thanks.

satishc1:
@Mpluya Sure.. please do. Thanks for the help.

Mpluya:
@satishc1 thank you.

satishc1:
@Mpluya Closed.

PR #1080, opened 2017-01-10
Dashboard name validation

Mpluya:
This is a follow up to https://github.com/capitalone/Hygieia/issues/1067.
Added client and server side validation on the dashboard name to only allow letter, numbers or spaces.


nireeshT:
@Mpluya  - Can we have the similar change for "rename dashboard" as well under same PR.

Mpluya:
@nireeshT good point. Will do.

Mpluya:
@nireeshT updates pushed. Summary of changes:

- added regex validation on rename of dashboard title
- enhanced server side validation on the required fields for a dashboard by adding a size check since UI posts empty string which bypasses the not null check
- added tests on rename dashboard

nireeshT:
@tabladrum @satishc1  - LGTM.

tabladrum:
LGTM

PR #1129, opened 2017-01-27
Spring Security/JWT Implementation

### LINKED ISSUE #1031: Authentication/Security Concerns
We have been looking into the security in Hygieia and have some concerns with how it’s implemented.  It looks like there is a call to the API service to login, but after that, all the authorization is in javascript and based on a couple cookies.  Security can be bypassed completely with very little effort.  Is there any work being done to address this? If not, we might look into improving this.

rob-miller-777:
Resolves #1031. This was considerably bigger than @brennanananan and I had realized going in, but we feel like we have it at a really good spot. Test it out and let us know if there are any changes you would like to see or issues we need to address.  We are anticipating plenty of questions due to the nature of the pull request and it's size. 

**What's Included**
- Implemented JWT to provide stateless authentication/authorization
- Updated angular client to work with JWT tokens
- Made token expiration configurable
- Implemented redirect functionality to take you to the page you were viewing if you token expires and the API returns a 401 status code
- Made secret key configurable to support having multiple instances of the API running
- Added Spring Security to Rest API
- Built in support for multiple forms of authentication (standard and ldap auth currently implemented)
- Refactored current login to work with Spring Security
- Created a UserInfo Repository separate from the Authentication repository to support multiple forms of authentication. This lets us control all application specific details of a user such as roles and themes independent of the authentication type. (Themes are still only handled by the angular app on the client, and saved in local storage, but this is something we can now save for a user if we want to)
- Made admin a "role" instead of just assigning admin privileges to the user named "admin". (Haven't implemented a way to assign new admins yet, so right now it works as before where the user named "admin" is given the role of admin, this is handled on the backend.  There is no way to give an ldap user "admin" privileges currently unless their name is actually "admin")
- Added method level security to dashboard modification controller so only owners of the dashboard or admins can modify a dashboard
- Added support for having multiple owners of a dashboard (have not created anyway to assign multiple owners yet, but shouldn't be hard to do)
- Implemented new owner functionality in such a way that it won't break backwards compatibility


Bradmoon:
i this pull request relates and solves issues outlined in #1031

MarkRx:
What's the state of this PR? Was hoping LDAP was in master by now but that does not appear to be the case.

rob-miller-777:
Haven't heard anything yet.

tabladrum:
@satishc1 lets branch this off

UI/src/app/dashboard/services/userService.js
```diff
@@ -0,0 +1,70 @@
...
+      this.hasDashboardConfigPermission = function (owner, owners) {
+    	if (this.isAdmin()) {
+    		return true;
+    	}
+
+    	var authtype = this.getAuthType();
+    	var username = this.getUsername();
+
+    	// preexisting dashboards
+      	if (authtype === 'STANDARD' && dashboard.owner === username) {
```
nireeshT: @rob-miller-777  - For preexisting dashboards there is no authType, this change will not provide admin access to the owner.

rob-miller-777: We changed the "Dashboard" class to include a list of "Owner" objects.  But we left the string field "owner" to maintain backwards compatibility.  So what this line does is says if the user is a "Standard" user, meaning they log in through Hygieia, and the owner is stored the old way and matches the current owner, let them modify.  The "authType" comes from the user, and not the dashboard object here.  Does that make sense?  Let me know if you need clarification.

rob-miller-777: Have you had issues modifying your previously created dashboards after running this branch?

nireeshT: @rob-miller-777  - yes, change dashboard.owner to owner because dashboard.owner is undefined in above line.
dashboard.owner is valued passed into as "owner" -> hasDashboardConfigPermission(owner,owners) and owner should be checked against username.

	if (authtype === 'STANDARD' && owner === username)

rob-miller-777: I see what you are talking about now, I'm blind, I'm guessing I was passing the entire dashboard through before, and changed it to just pass through the two different owners fields.  And forgot to update that.  Good catch.

rob-miller-777: I'll fix that now.

api/src/main/java/com/capitalone/dashboard/auth/AuthProperties.java
```diff
@@ -0,0 +1,70 @@
...
+	@PostConstruct
+	public void applyDefaultsIfNeeded() {
+		if (getSecret() == null) {
+			LOGGER.info("No JWT secret found in configuration, generating random secret by default.");
+			setSecret(UUID.randomUUID().toString().replace("-", ""));			
+		}
+		
+		if (getExpirationTime() == null) {
+			LOGGER.info("No JWT expiration time found in configuration, setting to one day.");
+			setExpirationTime((long) 1000*60*60*24);
```
nireeshT: can we set the expiration time only to LDAP login flow and not impact standard login? thoughts ?

UI/src/app/dashboard/services/userService.js
rob-miller-777: Fixed that, doesn't reflect here though.

api/src/main/java/com/capitalone/dashboard/auth/AuthProperties.java
rob-miller-777: I don't know if it's a good idea to create JWT tokens without expiration.  The expiration is really the only control the backend has to revoke the access to the user. Without an expiration, you could technically create a user, login, delete the user, and use the app forever with that deleted user. I'm assuming most users wouldn't notice the change since the only time you need to log in would be to make modifications to the dashboard, which technically shouldn't happen often after creating.  What concerns do you have with this approach of logging in daily by default?

nireeshT: @rob-miller-777  - if I set expiration time to 10 secs , i get logged off automatically after that expiration time. But as you mentioned only time user wanted to login is to edit widgets otherwise it is read-only. To maintain a better user experience can we set a lower threshold value as application timeout to 30 mins.

rob-miller-777: You can modify the expiration time to whatever you would like by setting this property.
`auth.expirationTime=[JWT expiration time in milliseconds]`
By default, it's set to 1 day if not provided.  Let me know if this didn't answer your question.

api/src/main/java/com/capitalone/dashboard/auth/AuthProperties.java
```diff
@@ -0,0 +1,70 @@
...
+
+@Component
+@EnableConfigurationProperties
+@ConfigurationProperties(prefix = "auth")
+public class AuthProperties {
+	
+	private static final Logger LOGGER = LoggerFactory.getLogger(AuthProperties.class);
+	
+	private Long expirationTime;
+	private String secret;
```
rob-miller-777: Correct, you have the ability to set the signing key for the JWT token, this gives you the ability to set the same signing key on multiple instances of the API in case you had multiple running for load balancing.  This would make sure that a request could be sent to any instance of the API and be validated properly with the key.

rob-miller-777: If you don't provide one, a random one would be generated, this works fine unless you have multiple instances of the API running.

nireeshT: @rob-miller-777 - Please update README.md

rob-miller-777: Will do.

rob-miller-777: Done, and updated with master.

api/src/main/java/com/capitalone/dashboard/auth/token/TokenAuthenticationServiceImpl.java
```diff
@@ -0,0 +1,90 @@
...
+		this.tokenAuthProperties = tokenAuthProperties;
+	}
+	
+	@Override
+	public void addAuthentication(HttpServletResponse response, Authentication authentication) {
+		String jwt = Jwts.builder().setSubject(authentication.getName())
+				.claim(DETAILS_CLAIM, authentication.getDetails())
+				.claim(ROLES_CLAIM, getRoles(authentication.getAuthorities()))
+				.setExpiration(new Date(System.currentTimeMillis() + tokenAuthProperties.getExpirationTime()))
+				.signWith(SignatureAlgorithm.HS512, tokenAuthProperties.getSecret()).compact();
```
nireeshT: @rob-miller-777  - we can add a layer of verifiability to JWT by adding Signature algorithm collection to Mongo with various signature algorithms and add specific algorithm id here or use setId() in Jwt. Later lookup with algorithmId and use the specific algorithm to get token with secret signing key.  This way we can mask the alg details and overcome security risks.

rob-miller-777: Just so we are on the same page, is your concern having the signature algorithm in the source code, potentially giving some additional information to malicious users?  Sounds like you would rather store that in the database, then pull the algorithm out by id.

rob-miller-777: If that is the case, we could add all the values of the enum to a collection in mongo with some id that makes sense.  Then add a configurable property to the API for selecting the algorithm you want.

rob-miller-777: I can add that functionality, but do you think it is a critical enough vulnerability that is required for this particular pull request, or could it be additional functionality we add later?

nireeshT: @rob-miller-777 - That's true, my opinion is that server should already know the algorithm and we can pull that out with id. As you mentioned we can add all enum to collection and look up from there. Lets get this change pushed in this PR if you think it is quick fix and we can get all changes merged at once.

nireeshT: @rob-miller-777  - this change is not critical to push it in this PR. We can have this functionality added later with a different PR.

rob-miller-777: Sounds good to me.

wdquan1985:
I test the ldap authentication function using this branch https://github.com/StateFarmIns/Hygieia/tree/springSecurityIntegration.
I use shell script to send build info to hygieia, for example:
curl --connect-timeout 10 -m 60 -i -X POST --header "Content-Type:application/json" -H "Accept:application/json" --data
'{"artifactName":"SANDBOX","canonicalName":"SANDBOX","artifactGroup":"SANDBOX","artifactVersion":"1.0","buildId":"timestamp","timestamp":1488942038335,"sourceChangeSet":[{"scmUrl":"http://127.0.0.1/gerrit/60730","scmBranch":"1.0","scmRevisionNumber":"Ica15fe4f1f9d05a3c6bf74b7c0df8191565fcf76","scmAuthor":"ZHANG,Harry(Harry)","scmParentRevisionNumbers":["d61cbc08f08bb8f674c41b4f55ff6e4fd784378d"],"type":"New","scmCommitTimestamp":1488963918000,"numberOfChanges":1}],"metadata":{"jobName":"SANDBOX","buildUrl":"127.0.0.1/job/SANDBOX/179/","buildNumber":179,"jobUrl":"http://127.0.0.1/job/SANDBOX/","instanceUrl":"http://127.0.0.1/"}}' -u userName:password http://127.0.0.1:8080/api/artifact .
I have send the username and password to hygieia, but there is an error:
{"timestamp":1489558957506,"status":401,"error":"Unauthorized","message":"Full authentication is required to access this resource","path":"/api/artifact"}

Then I permit the url request which start with api by changing code in the spring security java configuration file com.capitalone.dashboard.config.WebSecurityConfig.java.
@Override
protected void configure(HttpSecurity http) throws Exception {
http.headers().cacheControl();
http.csrf().disable()
.authorizeRequests().antMatchers("/appinfo").permitAll()
.antMatchers("/registerUser").permitAll()
.antMatchers("/login**").permitAll()
.antMatchers("/api**").permitAll()
.antMatchers(HttpMethod.GET, "/**").permitAll()
.anyRequest().authenticated()
.and()
.addFilterBefore(standardLoginRequestFilter(), UsernamePasswordAuthenticationFilter.class)
.addFilterBefore(ldapLoginRequestFilter(), UsernamePasswordAuthenticationFilter.class)
.addFilterBefore(jwtAuthenticationFilter, UsernamePasswordAuthenticationFilter.class)
.exceptionHandling().authenticationEntryPoint(new Http401AuthenticationEntryPoint("Authorization"));
}.

But it still can't work, how can I configure to ignore the authentication and authorization checks?
When I send the jenkins build into to hygieia, there is no any error, but the hygieia mongo db can't search the corresponding record.

wdquan1985:
I add .antMatchers("/build","/artifact","/quality/**").permitAll(), because I use shell script to send build info, artifact info and sonarQube report to hygieia server.
I have not used the jenkins plugin. I only use shell script, is your temporary solution avilable for shell script?
For example:
curl --connect-timeout 10 -m 60 -i -X POST --header "Content-Type:application/json" -H "Accept:application/json" -u userName:password --data '{~~~~~~}' http://127.0.0.1:8080/api/build
If I don't add .antMatchers("/build","/artifact","/quality/**").permitAll(), how can I make it work ?

rob-miller-777:
My most recent commit should allow your shell script to send data.  It pretty much does what you were doing.  Take a look at this commit here from last week. https://github.com/capitalone/Hygieia/pull/1129/commits/5cb339442d1c643aa8343ceaa4b212c347473841

wdquan1985:
Thanks, It can work now.

api/src/main/java/com/capitalone/dashboard/config/WebSecurityConfig.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+	protected void configure(AuthenticationManagerBuilder auth) throws Exception {
+		auth.authenticationProvider(standardAuthenticationProvider);
+		
+		String ldapServerUrl = authProperties.getLdapServerUrl();
+		String ldapUserDnPattern = authProperties.getLdapUserDnPattern();
+		if (StringUtils.isNotBlank(ldapServerUrl) && StringUtils.isNotBlank(ldapUserDnPattern)) {
+			auth.ldapAuthentication()
+			.userDnPatterns(ldapUserDnPattern)
+			.contextSource().url(ldapServerUrl);
```
nireeshT: @rob-miller-777  - DefaultLdapAuthoritiesPopulator sets a search scope of "ONE_LEVEL", but tosearch deep  "SUBSCOPE"  can we set a searchcontrols-> setSearchscope  to 2 or "SearchControls.SUBTREE_SCOPE" .

Path to searchScope - --->  this.AuthenticationManager -> providers-> LDAPAuthenticationProvider -> authoritiesPopulator-> ldapTemplate -> searchControls -> searchScope

rob-miller-777: I might be interpreting this incorrectly, but I don't see any reason to retrieve the authorities at all with this implementation.  All the role specific stuff, such as who is admin, is stored in the Hygieia database.  This would give the tool the ability to add new admins, instead of relying on ldap groups and such.

nireeshT: @rob-miller-777 - On the flip side , if we want to use the existing corporate ldap groups, is it possible to set this value?

rob-miller-777: I'm sure we could make it work, but it would require considerably more than just setting that value to integrate it with corporate ldap groups.  I would say that would be an additional pull request to do later.

nireeshT: @rob-miller-777  @tabladrum  @satishc1  Can we have a quick meeting regarding this next week to discuss on how we can integrate corporate LDAP groups.

rob-miller-777: It wouldn't be that difficult to use corporate ldap groups for controlling permission levels, as long as that was the only way to control admins for ldap users.  Could add properties to the props file for the ldap groups that should be considered admins, and send them through to a custom ldap authorities populator class.  Were you thinking something like that?  Control the access of "standard users" from the app, and "ldap users" using ldap groups?

rob-miller-777:
Where do we stand with this pull request?

satishc1:
@rob-miller-777 Should get merged next week.

api/src/main/java/com/capitalone/dashboard/auth/token/TokenAuthenticationServiceImpl.java
```diff
@@ -0,0 +1,90 @@
...
+		String authHeader = request.getHeader(AUTHORIZATION);
+		if (StringUtils.isBlank(authHeader)) return null;
+		
+		String token = StringUtils.removeStart(authHeader, AUTH_PREFIX_W_SPACE);
+		try {
+			Claims claims = Jwts.parser().setSigningKey(tokenAuthProperties.getSecret()).parseClaimsJws(token).getBody();
+			String username = claims.getSubject();
+			Collection<GrantedAuthority> authorities = getAuthorities(claims.get(ROLES_CLAIM, Collection.class));
+			PreAuthenticatedAuthenticationToken authentication = new PreAuthenticatedAuthenticationToken(username, null, authorities);
+			authentication.setDetails(claims.get(DETAILS_CLAIM));
```
satishc1: @rob-miller-777 Thanks for the contribution. I have a couple of questions on the JWT token.

The username that we get from the token... is it being checked against the Mongo user_info collection ?

The generation of the JWT token... is the assumption that it will be generated offline? Any chance we could provide an api to generate the token?

rob-miller-777: When you login, whether it be through Hygieia's auth, or LDAP, we take that username and get user info from the userinfo colletion such as roles, and put it in the token.  But it does not need to check against the collection on every request. 

I'm not sure what you mean by generated offline.  The token is generated upon login, and will be valid until the expiration time.

rob-miller-777: I imagine you could just use the login api call to get the token, and put it in your curl request.  As far as the jenkins plugin, I was thinking we could implement some sort of api key functionality instead of logging in.

satishc1: @rob-miller-777 Thanks.

satishc1:
@tabladrum @amitmawkin @nireeshT LGTM

xpdable:
@rob-miller-777 same as @wdquan1985 , I met the 401 even on the latest code base:

Could you hint where am I wrong? 
I've setup api- properties as with auth.secret=hygieiasecret

> [xpdable@xxxxxx01 ~]$ curl --connect-timeout 10 -m 60 -i -v -H 'Accept: application/json' -H Content-Type:application/json -X POST -d '{"executionId":"10","jobUrl":"http://jenkins.url.com","appName":"MYTESTAPP","envName":"MYTESTAPP","artifactName":"MYTESTAPP.jar","artifactVersion":"1.0.0-e693fe64-SNAPSHOT","jobName":"MYTESTAPP","instanceUrl":"MYTESTAPP.com","deployStatus":"SUCCESS","startTime":0,"artifactGroup":"MYTESTAPP","endTime":0,"duration":0,"startedBy":"MYTESTAPP"}'  http://10.35.24.62:3002/api/deploy/
* About to connect() to 10.35.24.62 port 3002 (#0)
*   Trying 10.35.24.62... connected
* Connected to 10.35.24.62 (10.35.24.62) port 3002 (#0)
> POST /api/deploy/ HTTP/1.1
> User-Agent: curl/7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.14.0.0 zlib/1.2.3 libidn/1.18 libssh2/1.4.2
> Host: 10.35.24.62:3002
> Accept: application/json
> Content-Type:application/json
> Content-Length: 341
>
< HTTP/1.1 401 Unauthorized
HTTP/1.1 401 Unauthorized
< server: Apache-Coyote/1.1
server: Apache-Coyote/1.1
< x-content-type-options: nosniff
x-content-type-options: nosniff
< x-xss-protection: 1; mode=block
x-xss-protection: 1; mode=block
< cache-control: no-cache, no-store, max-age=0, must-revalidate
cache-control: no-cache, no-store, max-age=0, must-revalidate
< pragma: no-cache
pragma: no-cache
< expires: 0
expires: 0
< x-frame-options: DENY
x-frame-options: DENY
< www-authenticate: Authorization
www-authenticate: Authorization
< content-type: application/json;charset=UTF-8
content-type: application/json;charset=UTF-8
< transfer-encoding: chunked
transfer-encoding: chunked
< date: Mon, 17 Apr 2017 06:58:09 GMT
date: Mon, 17 Apr 2017 06:58:09 GMT
< connection: close
connection: close

<
* Closing connection #0

rob-miller-777:
I'm having a hard time replicating this issue.  With the code base in master, the jenkins plugin doesn't seem to have any issue for me sending requests to the api.

rob-miller-777:
@wdquan1985 Are you still having this issue?

xpdable:
@rob-miller-777 Resolved.  The code allows 'http://10.35.24.62:3002/api/deploy' but not 'http://10.35.24.62:3002/api/deploy/' . Any slash at the tail is restricted :P

rob-miller-777:
Ah, makes sense.  Think that's ok though since the jenkins plugin doesn't add the slash.

xpdable:
@rob-miller-777 @satishc1 Regarding the possibility of AD LDS Service on LDAP, is there an plan of that ?

rob-miller-777:
I'm going to be perfectly honest and say I am not an LDAP expert, so I'm not sure what all would need to change.  However, from what little I do know, LDAP is just a protocol for querying something like active directory, so I would guess this should already be handled.  I could be misreading your question though.

xpdable:
@rob-miller-777 Honestly, I am not either but we'd like to integrate enterprise level authorization with that.
Let me explain a little more with the Jenkins setting. 
Our AD LDS is required a Manger DN and password for any user to login where I toggled in yellow background. Now for Hygieia setting I could only see the setting for the red underline part of user.
So questions is where I should put my Manager DN, etc...
Correct me if I am doing wrong in Hygieia API properties file :)

image

wdquan1985:
@xpdable - My ldap server info is as following:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
% ldapsearch \
-xLLL \
-h ************************.com \
-p 3000 \
-b "dc=Internal,dc=Users,dc=Bruce" \
"uid=bruce" \
uid displayName mail businessTitle JobFamily telephoneNumber mobile

dn: polarisuid=bruce,dc=Internal, dc=Users, dc=Bruce
uid: bruce
displayName: ******************
mail: ************************
businessTitle: ******************
JobFamily: ***********************
telephoneNumber: ***************
mobile: *********************
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

And I configure ldap for hygieia as following:
auth.expirationTime=
auth.ldapServerUrl=ldap://*******************.com:3791/dc=internal,dc=users,dc=Bruce
auth.ldapUserDnPattern=polarisuid={0}.

Then ldap authenticaition can work.

rob-miller-777:
With how it's implemented currently, there is no way to get the manager dn and password through, however, it should be a pretty trivial change to make this happen.  I am currently working on letting Hygieia utilize ldap groups to determine admin access, so I could include the ability to pass through more ldap settings such as manager dn.

rob-miller-777:
Is it not working for you with just those two settings available to you?

satishc1:
@rob-miller-777 @xpdable @wdquan1985 We will be merging a PR for ActiveDirectory support.

xpdable:
@rob-miller-777 problem is... 100 percent sure I need the Manager login to use our AD LDS,  therefore I am not sure where to set the password in those two settings even I could set DN after the ldaps url as @wdquan1985 does :sob: 

@wdquan1985 Have you even try some login with a admin and password in the config?

@satishc1  Thanks, anyway looks forward the AD support with manager credential

wdquan1985:
@xpdable - You don't need to set password in those two settings, you just need to enter you username and password in the login page.
login

rob-miller-777:
I believe it has to do with how it's configured.  I have a pull request I am going to open soon that allows you to set the manager dn and password.  It will also let you use an ldap group to select admins.

gokulknewt:
@wdquan1985,Team

We are going to implement the spring LDAP security for dashboard login page.Could you please provide me the steps and implementation.It would be much helpful!!

Regards,
Gokul k

wdquan1985:
@gokulknewt , which branch are you using? I use this one https://github.com/StateFarmIns/Hygieia/tree/springSecurityIntegration. Then you can configure the ldap authentication acorrding to my comments above.

gokulknewt:
@wdquan1985,

Thank you for your repsonse.I have imported ur code.For LDAP related what are all configuration i have to do the properties file.could you please guide me.

Regards,
Gokul k

gokulknewt:
@wdquan1985,

Also could you please let me know LDAP configuration in properties file.


Regards,
Gokul k

gokulknewt:
@Team,

Can we use tls 1.0 protocol for Spring LDAP integration with hygieia dasboard from properties file level ?

Regards,
Gokul k

wdquan1985:
@gokulknewt - Sorry for response so late because I have had a three day's vacation.
The following is my configuration in the 
And I configure ldap for hygieia as following:
auth.expirationTime=
auth.ldapServerUrl=ldap://*******************.com:3791/dc=internal,dc=users,dc=Bruce
auth.ldapUserDnPattern=polarisuid={0}
configuration

gokulknewt:
**@wdquan198**5-Thanks for your detailed info.Although,we have few clarifications.Can you please demonstrate on this.
**Q1#**There is seems to be u didn't define the LDAP username and password from applcation.properties file.Does it require to connect uname and pword?
**Q2#** auth.ldapServerUrl=ldap://*******************.com:3791/dc=internal,dc=users,dc=Bruce. 
       Can we use   tls 1.0 protocol instead of ldap protocols.
**Q3#** To authenticate the only admin user only can authenticate  others user also can we do the autnticate while entering login page.

Regards,
Gokul k

wdquan1985:
@gokulknewt - Q1# we don't need uname and pword in the application.properties file, because the uname and password will be captured in the login page where you try to login.
Q2# spring security framework supports ldap authentication, but I don't know if it support tls 1.0 authentication, you can search spring security materials to identify it.
Q3# I have used a commen user not admin user to authenticate, and it was successfull.

gokulknewt:
@wdquan1985,
we have tried to connect active directory,but while run the api.jar itself we are getting exception.Could you please let me know the configuration whether we configure right approach or not

**My configuration file looks like:**
auth.ldapServerUrl=ldap://dc2.qcorpaa.aa.com:389/OU=UD_DH_Nexus Rep Mgr,OU=Applications,OU=Groups,DC=qcorpaa,DC=aa,DC=com

auth.ldapUserDnPattern=Z1061076={0}

when run the api jar am getting exception like this
image

Regards,
Gokul k

rob-miller-777:
Try something more like this, 
`auth.ldapServerUrl=ldap://dc2.qcorpaa.aa.com:389`
`auth.ldapUserDnPattern=uid={0},ou=UD_DH_Nexus Rep Mgr,ou=Applications,ou=Groups,dc=qcorpaa,dc=aa,dc=com`

I'm not sure about the user dn pattern, you need the dn pattern to your users. And I don't know what yours would be.

gokulknewt:
@rob-miller-777 ,Thanks for your response.Still am getting same exception.
When am try to run the api.jar am getting exception.For that can you provide me the right direction.

Thanks,
Gokul k

rob-miller-777:
Think your issue might be with the spaces in "UD_DH_Nexus Rep Mgr", probably has to be url encoded. Try replacing that with "UD_DH_Nexus%20Rep%20Mgr".

wdquan1985:
@gokulknewt - Or you can try "auth.ldapServerUrl=ldap://dc2.qcorpaa.aa.com:389/DC=qcorpaa,DC=aa,DC=com", it's just a configuration issue based on your own ldap server, the hygieia code is OK.

gokulknewt:
@wdquan1985,

Could you please provide clear structure on auth.ldapUserDnPattern. Also to connect my Active directory i have define uname and password. how to define that.Could you please assist to us?

Here is the out dir. structure...
qcorpaa.aa.com/Support User Accounts/Applications/Devops dashboard,

Regards,
Gokul k

wdquan1985:
@gokulknewt - I want to help you but I think I had told you all the things I know. May be you can setup a simple ldap server for verifying the ldap authentication. Or you can capture the data package using tcpdump command in Linux or wireshark in windows.

Regards,
Bruce

Issue #1103, opened 2017-01-19
Cross-origin resource sharing error between api and ui


jshpopelka272:
I'm trying to get Hygieia running in containers with Openshift. 

I have three separate pods: mongo, api, ui. The api layer is able to communicate with the mongo pod. 

The UI layer is running, but not able to connect to the api layer. In my Chrome browser console, I see this error:

> XMLHttpRequest cannot load http://api-hygieia.apps.cluster.local/api/dashboard. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://ui-hygieia.apps.cluster.local' is therefore not allowed access.

When I add the cross-origin resource sharing extension to my browser, they are able to connect. 

Is there a better solution for this so all users won't have to add the extension?

tabladrum:
@amitmawkin please see.

jshpopelka272:
Update: I was able to find a solution to this CORS filter issue.

By adding these four lines of code into the LoggingFilter.java file and running the maven build on the api layer, cross origin domains are accepted.

`httpServletResponse.addHeader("Access-Control-Allow-Origin", "*");
httpServletResponse.addHeader("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE");			httpServletResponse.addHeader("Access-Control-Allow-Headers", "Content-Type");
httpServletResponse.addHeader("Access-Control-Max-Age", "1");`

I'll be adding a pull request for this so others can avoid this issue.

PR #1134, opened 2017-01-30
Update LoggingFilter.java

jshpopelka272:
Fixes issue with Cross Origin Resource Sharing between api and ui layer.


satishc1:
resolves https://github.com/capitalone/Hygieia/issues/1103

satishc1:
@tabladrum Should this be parameter based ?  By default it could be off. If a user decides they want to enable CORS they can choose to do it.

satishc1:
@amitmawkin Any concerns on this one ?

tabladrum:
@satishc1 parameter based sounds good.

api/src/main/java/com/capitalone/dashboard/config/LoggingFilter.java
```diff
@@ -108,6 +108,13 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             }
 
         } else {
+            //new code added 
+			//adds headers to response to allow CORS	
+			httpServletResponse.addHeader("Access-Control-Allow-Origin", "*");
```
nireeshT: @tabladrum @satishc1 @amitmawkin  - should we be allowing any origin to access Hygieia.By using * in Access-Control-Allow-Origin, we enabled every application to  send a XHR request to Hygieia and I think we should have a control/logging of calling application by creating a collection of whitelisted urls and enabling them in "Access-Control-Allow-Origin" by replacing * with listOfUrls. 

Eg :  "Access-Control-Allow-Origin":"*"
Replaced with : "Access-Control-Allow-Origin":  listOfUrls 

where listOfUrls - "http://google.com,http://dummy.com,....etc"

That way we have access to enable to particular resources instead of all.

api/src/main/java/com/capitalone/dashboard/config/LoggingFilter.java
```diff
@@ -108,6 +108,13 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+			
+			httpServletResponse.addHeader("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE");
```
nireeshT: @tabladrum @satishc1 @amitmawkin  - methods GET and POST are valid.

api/src/main/java/com/capitalone/dashboard/config/LoggingFilter.java
MarkRx: Allowing arbitrary callers is a potential security vulnerability. It should be restricted to only those that are needed.

PR #1149, opened 2017-02-03
Gitlab Bug Fix

rob-miller-777:
We found a bug in the gitlab collector that would stop you from being able to get commits if the collector ran for a repo, but the collector was misconfigured.  What would happen is the collector would have a bad configuration that didn't allow you to retrieve commits from git.  If someone then created a Gitlab repo in a dashboard, the collector would run, retrieve no commits, and set the last updated to the current time.  The "first run" would have been run already.  Then, when the configuration was fixed, the collector would have already done the first run, so it would only try to retrieve commits from the "last update" time, back 10 minutes.  Therefor never retrieving the commits from whatever value was specified for first run.  This change prevents updating the "last update" time on a repo unless there was a successful request made to gitlab.  Any HttpClientErrorException or ResourceAccessException will prevent the "last update" time from changing, which means, a newly configured repo will get it's first run again, and other repos will get all commits from the last successful run.  

Sorry for the long explanation, I think there is probably more explanation than actual code.  Let me know if you have questions.


satishc1:
LGTM

PR #1182, opened 2017-02-17
uDeploy token auth

satishc1:
Add ability for token based auth in uDeploy collector.


satishc1:
@tabladrum @amitmawkin LGTM

PR #1185, opened 2017-02-17
Correct logback on gitlab-git-collector

Mpluya:
Adjusted the logback configuration on the gitlab-git-collector to keep it consistent with the other Hygieia collectors.


satishc1:
@tabladrum LGTM

PR #1211, opened 2017-02-27
Skip frontend plugin

rob-miller-777:
Adding an optional parameter to allow us to skip running the frontend maven plugin if desired.  By default, it still runs as it did.


satishc1:
@tabladrum LGTM

Issue #1136, opened 2017-01-31
Password / Documentation


benjaminwootton:
I'm having problems getting the frontend and API talking due to the password.

Master API contains:

[centos@devopsmetrics0 Hygieia]$ more api/target/classes/application.properties
dbname=dashboarddb
dbusername=dashboarduser
dbpassword=ENC(C9PLVDJ/1OfkT/3jvC/wEB20R1XPJtg9)
dbhost=localhost
dbport=27017
dbreplicaset=false
dbhostport=localhost:27017
server.contextPath=/api
server.port=8080

But the documentation says to setup Mongo with db / dbpass as the user.

I've tried reverse engineering the encrypted password, or setting a new password in Mongo, but the Java stuff for creating an encrypted password hangs.

Any clear steps on how to get the two talking, or can we update the repo so it's easier to get started?

benjaminwootton:
If I do a clean checkout, docker build and docker-compose run, api service fails with this:

aused by: java.lang.IllegalArgumentException: Password cannot be set empty
	at org.jasypt.commons.CommonUtils.validateIsTrue(CommonUtils.java:150)
	at org.jasypt.encryption.pbe.StandardPBEByteEncryptor.resolveConfigurationPassword(StandardPBEByteEncryptor.java:792)

I think because JASYPT_ENCRYPTOR_SECRET is not set.  There isnt much mention of this in the docs either.

benjaminwootton:
Related - https://github.com/capitalone/Hygieia/issues/796

satishc1:
@benjaminwootton 
Refer to api module readme updated by @Mpluya 
https://github.com/capitalone/Hygieia/tree/master/api

Via docker, pass as an environment variable docker run -t -p 8080:8080 -v ./logs:/hygieia/logs -e "SPRING_DATA_MONGODB_HOST=127.0.0.1" -e "JASYPT_ENCRYPTOR_SECRET=hygieiasecret" -i hygieia-api:latest then pass the environment variable as a Spring boot system property java -Djasypt.encryptor.password=$JASYPT_ENCRYPTOR_SECRET -jar api.jar --spring.config.location=dashboard.properties

Mpluya:
[...]

PR #1217, opened 2017-03-01
removed the use of environment variable JASYPT_ENCRYPTOR_SECRET as th…

Mpluya:
This is a fix to a previous pull request around encrypting properties. The change is to move away from using an unnecessary environment variable that added a layer of indirection in favor of just utilizing the environment variable automatically picked up by jasypt-spring-boot-starter.


Mpluya:
This is an action item I took from https://github.com/capitalone/Hygieia/issues/1136.

MarkRx:
Please also remove the ENV variables set in the api dockerfile

satishc1:
@tabladrum LGTM

PR #1250, opened 2017-03-17
Allow use of Docker Config JSON for private registry

brennanananan:
This PR adds a property to the main project POM to allow the docker maven plugin to use the host's ~/.docker/config.json file to authenticate to private registries. If none is specified when building images with maven, default value is false to maintain transparent backwards compatibility.


satishc1:
@tabladrum LGTM

Issue #1247, opened 2017-03-16
Question regarding Version One Collector


brennanananan:
We're trying to deploy an instance of the version one collector with Docker, so on container creation the versionone-properties-builder.sh is applying properties for the collector jar. I noticed that many of the other properties builder scripts are applying a CRON setting 

```bash
# from multiple files...
jenkins.cron=${JENKINS_CRON:-0 0/5 * * * *}
sonar.cron=${SONAR_CRON:-0 0/5 * * * *}
subversion.cron=${SUBVERSION_CRON:-0 0/5 * * * *}
```

However, the versionone builder script has no such property being set. Is this intentional? Am I missing something?

satishc1:
Closed.

PR #1251, opened 2017-03-17
Fixed issues in version one properties builder shell file

brennanananan:
When using docker to create a versionone collector, we found that there was no properties file was being created. This was due to a syntax error in the properties builder script when copying the values to the properties file. Once that file was created, the was no CRON being specified. CRON has been added to the versionone properties builder as well as the syntax errors being corrected.


satishc1:
@tabladrum LGTM

satishc1:
resolves https://github.com/capitalone/Hygieia/issues/1247
@brennanananan Thanks.

PR #1272, opened 2017-03-27
Fix for issue #1265

MarkRx:
Calls to check if optional is set need to use isPresent as get will throw an exception if it is not set.

#1265


satishc1:
@tabladrum LGTM

PR #1275, opened 2017-03-28
Fixing mongodb and hygieia api configuration

jdewinne:


docker-compose.yml
```diff
@@ -1,21 +1,24 @@
 mongodb:
-  image: mongo:latest
-  container_name: mongodb
-  command: mongod --smallfiles
-  ports:
-   - "27017:27017"
-  volumes:
-   - ./mongo:/data/db:rw
-  volume_driver: local
+  image: bitnami/mongodb:latest
```
ebosveld: Why did you change this to a Bitnami image over the official MongoDB image? What is the value of that change?

jdewinne: Sorry for the late reply. I think it had to do with https://github.com/docker-library/mongo/issues/158
But that might now be fixed on the monmgodb side.

satishc1:
@tabladrum @jdewinne Replaced the bitnami with the standard image.

PR #1277, opened 2017-03-28
Docker Fix for Gitlab Feature Collector

rob-miller-777:
Had the wrong filename, so wasn't taking the properties.


satishc1:
@tabladrum LGTM

PR #1279, opened 2017-03-29
Gitlab feature collector - optional server port

brennanananan:
Currently the Gitlab feature collector accepts a Gitlab server with a corresponding port in the properties builder file. If the collector is started without a Gitlab port supplied, the port is processed as "", throwing a Number Format Exception. The logic in the URI builder has been updated to not include the port if it is an empty string.


satishc1:
@tabladrum LGTM

PR #1286, opened 2017-03-30
Gitlab SCM Cleanup

rob-miller-777:
Only functionality change is making the port optional if running in a docker build.  Before an empty string would be used if the port was not provided, causing an integer parse exception.  Now it won't try to add the port if it is blank.  

Other changes are just to make the code more consistent with other collectors and minor refactoring.


rob-miller-777:
Can we kick the Travis build off again, had an encryption issue that wasn't related to my code changes.

satishc1:
@rob-miller-777 I've restarted the build.

satishc1:
@tabladrum LGTM

PR #1288, opened 2017-03-30
Fix for Rundeck - pattern match regex for default options

nireeshT:
Fix for Rundeck - pattern match regex for default options
made artifact version mandatory.


satishc1:
@tabladrum LGTM

PR #1292, opened 2017-04-04
Monitor widget updates

### LINKED ISSUE #986: Monitor Widget should handle authentication for services
This is to handle 403's that result in a visual that indicates service is down.

### LINKED ISSUE #1224: Provide Proxy support for Monitoring Widget.

brennanananan:
This PR will resolve #986, and will resolve #1224 

This PR enhances the monitor widget in three different ways:
* **UI Side**
  * It adds a new icon to denote services explicitly as unauthorized (this icon is a lock, and it is green due to the fact that in order to get the unauthorized the service must be on and running)
  * It adds a legend for quick explanations of what the different icons mean
* **API Side**
  * It adds support to apply proxy credentials via the API properties file to allow the ServiceService to ping endpoints outside of the proxy.

**Also Included**
* Slight refactor of how service health checks are made to make it more testable.
* Dashboard ID is passed to layer where call is being made - this ensures more informative logging around which dashboard is having connectivity issues and/or which dashboard is pinging endpoints it possibly should not be pinging.


core/src/main/java/com/capitalone/dashboard/model/ServiceStatus.java
```diff
@@ -4,14 +4,36 @@
  * Denotes the status of a product or service
  */
 public enum ServiceStatus {
-    Ok, Warning, Alert;
+    Ok(200, 200),
+    Warning(300, 400),
+    Unauth(401, 401),
```
nireeshT: @brennanananan  - why not 401 and 403 here.

brennanananan: @nireeshT If we have a status for 403, this insinuates that credentials are being passed to the monitored service. At this time, no credentials are being passed, so 403 was not included.

nireeshT: @brennanananan  - which case I see 401 error, can you put one scenario. I am trying to recreate the UNAUTH case and no luck !!.

brennanananan: @nireeshT @rob-miller-777 This url should return a 401 from GitHub: https://api.github.com/?access_token=OAUTH-TOKEN

satishc1:
@tabladrum LGTM

PR #1296, opened 2017-04-05
Gitlab optional path

### LINKED ISSUE #1291: Running Gitlab behind Reverse Proxy
I'm running my Gitlab behind Nginx proxy. So, I'm using the context "/gitlab". But there's no way to configure Gitlab Collector to look to this context. The only possible configuration is to point directly to host:port.

rob-miller-777:
Resolves #1291 Added an optional path parameter in case it is needed to access the instance of gitlab.


satishc1:
@rob-miller-777 Any change needed in the docker properties ?

rob-miller-777:
Yep, knew I was forgetting something.

satishc1:
@tabladrum LGTM

PR #1297, opened 2017-04-05
Change load order in AWSCredentialsProviderChain.

satishc1:
Fix problem where ProfileCredentialsProvider is not getting loaded because InstanceProfileCredentialsProvider is always loaded first and is mapped to the default role linked to the EC2 instance.


satishc1:
@tabladrum @amitmawkin LGTM

amitmawkin:
LGTM

PR #1338, opened 2017-04-21
Gitlab Ignore Untrusted Certs/Log Location fix

rob-miller-777:
Updated the feature collector to function like the scm collector in being able to trust self signed certificates.  I also changed the scm collector to only set the property once, instead of on every request.  There is also a small logging file fix that puts the log files in a log folder like the other collectors.


satishc1:
@tabladrum LGTM

Issue #1307, opened 2017-04-07
Hygieia - Conference Call


amitmawkin:
Calling all folks who are actively participating in Hygieia development. Please see discussion topic below on the agenda and please feel free to add more if needed.

Topics

- Hygieia Github Code Base restructuring
- Hygieia Build optimization

@tabladrum & I will facilitate with dates/logistics etc.

If you would like to be included please tag yourself to this thread.

amitmawkin:
Tagging some known contributors.
@MarkRx 
@rob-miller-777 
@jaredsmith 
@brennanananan 
@swizzley

rob-miller-777:
#1293 Which might go hand in hand with restructuring the code base.

alagus84:
@alagus84

amitmawkin:
We are still working on the logistics and approvals from our end. Just wanted to check on everyone's willingness to attend the call. Please respond to this thread

rob-miller-777:
@brennanananan and I are planning on attending.

alagus84:
@alagus84 adding myself. Yes interested

swizzley:
@amitmawkin ready and willing

xpdable:
@xpdable Shall I ? We've got some resource and willing to contribute to the project

satishc1:
@xpdable Absolutely.

gttps:
@gttps

anjana83:
@anjana83

sibaom:
@sibaom interested

amitmawkin:
thanks all , will be sending the date and time for this shortly. Thanks for showing the interest.

amitmawkin:
Guys here are the logistics details for the meetings

May 25 2PM EST. 

https://hangouts.google.com/call/ej44lzyjbrfjjoc7mf6he5jmrqu

amitmawkin:
Talk to you all tomorrow. Please join and discuss the features you all want in Hygieia and help  shape up the roadmap

amitmawkin:
https://hangouts.google.com/call/ej44lzyjbrfjjoc7mf6he5jmrqu

amitmawkin:
Talk to all of you at 2pm E.S.T today please do try to attend

https://hangouts.google.com/call/ej44lzyjbrfjjoc7mf6he5jmrqu

rob-miller-777:
Didn't catch at the end how you were going to send out that roadmap.

tabladrum:
[...]

PR #1340, opened 2017-04-24
Ldap admin group

### LINKED ISSUE #1326: API docker properties-builder.sh is missing authentication properties
A number of properties are missing from the docker properties builder.

https://github.com/capitalone/Hygieia/blob/master/api/docker/properties-builder.sh

> auth.expirationTime=[JWT expiration time in milliseconds]
> auth.secret=[Secret Key used to validate the JWT tokens]
> auth.ldapServerUrl=[LDAP Server Url, including port of your LDAP server]
> auth.ldapUserDnPattern=[LDAP User Dn Pattern, where the username is replaced with '{0}']

@rob-miller-777 @nireeshT

rob-miller-777:
@nireeshT had brought up using LDAP groups to control admin users in Hygieia in #1129 .  After discussing with our team, we thought this would be a better approach as well.  This pull request would take control of admins for LDAP users outside of Hygieia, and utilize LDAP groups instead.  You will now be able to set an LDAP group as a configuration property, that contains all of Hygieia admins.  So if I have an LDAP group named "hygieia_admins", I would set `auth.ldapAdminGroup=hygieia_admins`.

To make this functionality work, I moved the "roles" property from the UserInfo collection, to the Authentication collection.  This means that currently, the UserInfo collection doesn't really do anything, but in the future it could be used to save application specific configurations such as themes.  

Admin control for standard users would still need to be controlled within the app, and there is a pull request coming for that functionality shortly.  

Resolves #1326 as well.


MarkRx:
I'm not sure about this one. If you look at other tools they all have their own internal permissions model. Some do hook up to LDAP/AD but that is optional. This provides ease of use and flexibility. The problem with using LDAP/AD for permissions is that in a large corporation there could be a lot of overhead and processes in order to add/remove/manage users in LDAP/AD groups. This solution would not work for us.

rob-miller-777:
I can see both sides, initially we developed it so all permissions would be controlled within the app, but there were requests to control admins using ldap groups.  Possibly another talking point for #1307

rob-miller-777:
I could also do something similar to what you were saying where you can choose between app control and ldap control, but it could get a little ugly.

satishc1:
@rob-miller-777 @MarkRx @tabladrum 

I agree with MarkRk on the overhead piece.

Also we wanted to use the user_info collection to store additional attributes from ldap such as email address, full name etc.

Is there a reason why you removed the record that was getting created in the user_info collection when a standard or ldap user signs in  ... and can you revert this change?

rob-miller-777:
If you guys don't like the idea of controlling admins through ldap groups, we can kill the merge request.  I have another branch that let's us modify admins from within the app for both standard and admin users that I will open a pull request for.

satishc1:
@rob-miller-777 Yes... I think that would be better. Please revert this PR. I haven't looked at the other PR yet.. https://github.com/capitalone/Hygieia/pull/1342 .  Does 1342 relate to this one?

rob-miller-777:
#1342 is built on this one for adding new admins for standard users.  #1354 manages both standard and ldap users from within the app, so you will probably like that one more.

PR #1354, opened 2017-04-27
Standard/LDAP Admin Management

rob-miller-777:
@satishc1 @MarkRx @tabladrum

This pull request could be used in place of #1340 and #1342 , this lets you manage both standard and ldap users from within the application.  I had initially created this one before seeing what it would take to utilize ldap groups.  See what you think.


MarkRx:
Could you add your changes to api/docker/properties-builder.sh to this?

rob-miller-777:
Will do

satishc1:
@rob-miller-777 This looks great. Thank you for the contribution 👍 

@rob-miller-777 @MarkRx @tabladrum I noticed a few things..

- How do we create the initial admin user? Just sign up ? Create in mongo ?

- The user_info collection doesn't specify that the admin user has ROLE_ADMIN in its authorities. How is the app identifying that the admin user is indeed an admin without having the role?

- On the Manage Admins screen, looks like it doesn't let you remove the last entry from the right hand side. And if I add and remove the "admin" user, does it affect anything?

rob-miller-777:
Yeah, all these issues are related and are based on your first point, how do you create the first admin user?  To have an initial admin user, there is some logic in there that ensures the "standard" user with username "admin"  is always admin. [See here](https://github.com/capitalone/Hygieia/pull/1354/files#diff-714ec0af3e35cc9d1dd046a3e82efbb4)  This is a hack, and should be replaced by something better, but I didn't want to tackle that in this merge request, and it's how it currently works, so wasn't really changing anything.  So I do agree this needs to be addressed, just not sure how to handle it yet.

As far as removing admins, I don't allow you to remove the last admin because then you would be in trouble, and currently, removing the admin privilege from the "standard" user named "admin" won't have any effect as that user will always have admin privileges.

satishc1:
@rob-miller-777 
I'm ok with hard-coding admin to be the admin user. This could be setup feature where we add admin during the initial install.

The add/removing admins screen is not working right. What I'm seeing is I can add the users in any order. But while removing, it seems to be hard-coded to stop when there's one entry left in the list..regardless of who that last user might be.

We should probably exclude the "current" user from the LHS (or disable the button for the current user).

rob-miller-777:
That is working as intended, there is some logic in the UI that hides the button if there is only one admin user left.

As far as disabling the button for current user, do you not want admins to be able to modify their own privilege?

satishc1:
@rob-miller-777 I was not concerned with admins being able to modify their own privilege. My concern was the admin can't remove the last user from the RHS. But playing around with it, I can see there is a workaround. Add the "admin" to the RHS, then remove the rest.

satishc1:
@tabladrum LGTM

rob-miller-777:
Exactly, and once we figure something out so the "admin" user isn't always admin, it will make more sense.

PR #1356, opened 2017-05-01
(Performance Fix) Change deploy widget to use uib-typeahead

MarkRx:
This widget was skipped when changes were made to other widgets to convert them to use the uib-typeahead directive (#1049). This PR changes the deploy widget to use the typeahead mechanism. It works with or without the "aggregateServer" option. I also fixed an XLD bug while in here.

When aggregate servers is on it obtains data using the following algorithm:
```
        /*
         * Obtains deployment jobs using recursion when necessary. 
         * 
         * It is necessary to make additional calls when 'aggregateServers' is true since we need all like applications on different servers 
         * to be available when saving our data. To do this we compare the last item in the list from our first paged call to all subsequent data.
         * If the application name and id is the same then we keep the data in our list. Once we encounter data that is different we can cease recurision
         * since the calls are sorted.
         * 
         * Example:
         * Suppose a size of 3 is used with aggregate servers and our first REST call returns the following:
         *   Deployment A (http://deploy.instance1.com)
         *   Deployment A (http://deploy.instance2.com)
         *   Deployment B (http://deploy.instance1.com)
         *   
         * We need to make an additional rest call to see if there are any more 'Deployment B' jobs. Suppose the second REST call returns the following:
         *   Deployment B (http://deploy.instance2.com)
         *   Deployment B (http://deploy.instance3.com)
         *   Deployment C (http://deploy.instance1.com)
         *   
         * We will keep the Deployment B's returned from the REST call and ignore everything that comes after it. Since the last item in this list is different
         * than the name + id (not shown) for our original REST call we cease searching for more items.
         */
```

### DIFF
--- a/UI/src/components/widgets/deploy/config.html
+++ b/UI/src/components/widgets/deploy/config.html
@@ -3,29 +3,22 @@
           ng-submit="deployConfig.submit(configForm.$valid,deployConfig.deployJob)" novalidate="novalidate">
         <form-group input="deployJob" errors="{required:'Please select your application'}">
 
-            <!--<select name="deployJob"
-                    class="form-control"
-                    ng-model="deployConfig.deployJob"
-                    ng-options="job.name for job in deployConfig.deployJobs | orderBy: 'name' track by job.value"
-                    ng-disabled="deployConfig.jobDropdownDisabled"
-                    required>
-                <option value="">{{deployConfig.jobDropdownPlaceholder}}</option>
-            </select>-->
-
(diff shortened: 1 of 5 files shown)


satishc1:
@MarkRx @tabladrum LGTM. 👍  The jira and v1 feature widgets project and team dropdowns also need a similar fix... just saying :)

MarkRx:
True though in those cases the data returned should be reasonable (a few hundred at most) :)

PR #1365, opened 2017-05-04
Configurable Authentication Providers

rob-miller-777:
This pull request would make the authentication providers configurable.  By default, the only authentication provider would be the "Standard" hygieia authentication.  You can also choose to have both LDAP and Standard, or just LDAP.  This was created because we wanted to only give our users the choice of LDAP authentication.  It is also useful for organizations that don't utilize ldap because you can just hide that option from being shown.  Not only does this pull request hide these authentication methods on the UI, but also prevents them from being used to authenticate on the API as well.


satishc1:
@rob-miller-777 @tabladrum Awesome 👍

satishc1:
@rob-miller-777 Please rebase. Couple of merge conflicts.

rob-miller-777:
Done, thanks @brennanananan !!!

satishc1:
@rob-miller-777 @brennanananan Cool thanks. I was just gonna mention the colors seemed off. I'll pull the changes down.

brennanananan:
@satishc1 There were a couple issues introduced by me merging, they should be fixed now.

satishc1:
This is what was there before
image

This is what i see now with this PR
image

The previous look seems more clear as to which tab is currently selected. What do you think ? @brennanananan

rob-miller-777:
That change isn't from this pull request. It came from another request a couple weeks ago I believe.  I'm cool with it either way, I think it's evident enough with just the white outline which one you are on.

satishc1:
@tabladrum LGTM

PR #1374, opened 2017-05-08
Active directory support and administer dashboard owners.

satishc1:
This PR adds AD support. It also allows a dashboard to be owned by multiple users. This will enable standard users to assign their dashboard to their ldap id. The process for this will be to login to hygieia with your ldap id. This will create an entry in the user_info collection. Then login to hygieia with your standard id. Then assign the ldap id as an owner of your dashboard. Eventually you'll be able to turn off standard authentication and migrate hygieia to ldap.


satishc1:
@tabladrum LGTM

xpdable:
@rob-miller-777 @satishc1  Regarding the connection to our AD with Admin user/password setting, this does not work for that, does it?
Where I mean the setting like we did in Jenkins.
image

satishc1:
@xpdable You shouldn't need an admin user/password setting. Try it setting it up based on the auth.AD properties in https://github.com/capitalone/Hygieia/blob/master/api/README.md

rob-miller-777:
On the "Administer your dashboard" modal, I think it would be a good idea to change the wording from "Admin" to "Owner".  I got confused and thought I was controlling admins of the application itself.

satishc1:
@xpdable I thought about that... problem is ..on the RHS it lists the Dashboard owners and the Administrators of hygieia (who many not own that dashboard directly).

rob-miller-777:
In my opinion, I don't think admins should be shown on the right side.  Only the people who are actual owners of the dashboard.  You probably shouldn't know, or care who the admins of the application are.

rob-miller-777:
It will also get confusing if you try to remove an admin, because they will just keep going back to being an owner on refresh.  It would probably be a lot simpler to just show the users who are listed as owner of the dash on that modal.

satishc1:
@rob-miller-777 Let us think about this a little more... our initial thought was to list owners and the administrators on the RHS. The main use of this screen was to enable people to migrate their dashboards from the standard login to their ldap login. We will get feedback from people here as they start to use the screen.

I was also debating whether there is anyway to integrate the "Admin" screen that you created with this "Dashboard Admin" dialog ... off course currently there is no way to get to the "Admin" screen for non-admin users.

rob-miller-777:
I think we should probably clarify the difference between an owner and an admin.  My thoughts would be an owner of a dashboard should have the ability to modify and delete the dashboards they are owners of.  An admin would have the same ability for every dashboard in the application, as well as controlling who other admins are, and other functionality we may want to give them in the future.  

Because of this, you would not want to integrate the admin page with the dashboard owner dialog, because they are completely different concepts.  The dashboard configuration dialog, in my opinion, should only be used to view and modify other owners of the dashboard you are looking at, having the admins incorporated as well becomes confusing because you really can't remove an admin from the RHS.

xpdable:
@satishc1 @rob-miller-777  Both, sorry for confused the 'Admin'.  While I say 'admin', I mean the manger DN from our AD LDS service. We definitely need a manager authorization for the connection to our AD LDS service (refer to the Jenkins setting). There are many restriction settings apart from the manager DN, like we also need to set some 'Group Search Base', 'Group Search Filter' etc...

The process is like, when initially the connection to AD LDS is established, we need a manager DN with password to access the content where call a bind action. After successfully bind the manger account, we could then see the DN that we have....   Without bind the manager DN account, nothing is explored in the AD LDS servise.

Some more info pasted from ldp.exe console:

> ld = ldap_sslinit("XXX.XXX.XXXX.COM", 50001, 1);
Host supports SSL, SSL cipher strength = 128 bits
Established connection to XXX.XX.XXXX.COM.
Retrieving base DSA information...
Result <0>: (null)
Matched DNs: 
Getting 1 entries:
>> Dn: 
	1> currentTime: 5/11/2017 9:48:32 China Standard Time China Daylight Time; 
	......
-----------
res = ldap_bind_s(ld, NULL, &NtAuthIdentity, 1158); // v.3
	{NtAuthIdentity: User='s.xxx.xxxx'; Pwd= <unavailable>; domain = 'TEN'.}
Authenticated as dn:'s.xxx.xxxx'.

rob-miller-777:
You currently can't pass those manager credentials in, however, it would take much to make that possible. There are two different issues being discussed here, the UI for adding new owners, and being able to integrate with AD using manager credentials.

xpdable:
@rob-miller-777 Thanks. I am talking about second issue.

Issue #1339, opened 2017-04-23
Handle bad credentials for scm widget


satishc1:
Invalid credentials should be handled gracefully by the github collector and displayed on the dashboard. Currently the collector keeps polling with invalid credentials.

rob-miller-777:
Was thinking about this when working on the gitlab collector, need some way to show the user that there is an error, just not sure how to do it with what's currently available.

nireeshT:
@satishc1 @rob-miller-777  - Also we may need to halt collectors when there is an error. My opinion,  we have to set a flag(or a value) to mongo and check for collector status. Use the same flag to propagate as an error in UI, code should handle to catch exception and terminate the collectors when occurred.  Thoughts?

tabladrum:
Here is my design suggestion:
Set a flag in collectorItem indicating a potential bad "url" (including bad credential). Set it to true when encounter non 200 response multiple times (configurable via properties). Unset only via UI. Also, the UI widget should indicate that the "url" is encountering trouble.

tabladrum:
Potentially can set "enabled" flag on the collector to false too. Only to be modified from UI.

rob-miller-777:
I don't know about halting the collector, at least with github/gitlab, you can get errors when the user misconfigures there repo, or using a private repo without providing credentials.  This would then stop all repos from being collected.  

My suggestion:
Create a new collection for "Dashboard Errors", which can be populated by any of the collectors.  Then add something in the UI for displaying the dashboard errors to the user.  There are probably other cases where things fail silently that we should be displaying to the users.

tabladrum:
While you are at it - fix one more bug too: 
The SCM widget creates duplicate collector items. Can you fix that too :-)

satishc1:
@rob-miller-777 The collector should stop polling a configured scm (collectorItem) as soon as it encounters an error. We could relax this rule if we can correctly identify the type of error. For example a network error is not reason enough to stop polling, but bad credentials or bad url is reason enough.

rob-miller-777:
That would be easy enough for Gitlab, as both bad credentials, and bad url would result in a 404.  Even with that, it would still be worthwhile to alert the user so they can fix the problem.

satishc1:
@rob-miller-777 Definitely. We can read the flag on the collectorItem and update the status of the widget. The widget currently displays the time when the last poll happened. @nireeshT is familiar with this piece. The status could be right next to it.

tabladrum:
[...]

PR #1388, opened 2017-05-15
Fix for Issue# 1339

tabladrum:
Changes:
1. Collector Item and Collector has CollectionErrors. Collectors must set it by catching exceptions.
2. After a certain threshold (configurable in settings) collector stops collecting for the collector item.
3. UI widget indicates if collector item has errors. Errors are cleared when the widget config is saved again
4. Also, fixed an issue with SCM collector item creation - every time there was a "save" config, it was creating a new collector item. Needed to create some custom repository query to find collector item matches via partial list of options. 

Resolves https://github.com/capitalone/Hygieia/issues/1339


satishc1:
@tabladrum LGTM

PR #1402, opened 2017-05-18
Changes to allow other security analysis tool to show stats.

tabladrum:
Each tool has its own way of categorizing issues such as "Critical", "Blocker" etc and then some tool has "High", "Medium", "Low" etc. This change will display whatever metric is collected by simply using the name, value pair.


satishc1:
@tabladrum LGTM

PR #1413, opened 2017-05-23
Feature widget: uib-typeahead implementation for Gitlab and VersionOne

nireeshT:
uib-typeahead implementation for Gitlab and VersionOne


satishc1:
@tabladrum LGTM

PR #1420, opened 2017-05-25
Splunk logger

rob-miller-777:
Added an optional filter in the API that will log connection info in a format that is easily consumable by Splunk.  This will help to debug the application in the environment, and provide useful metrics around usage.  Also refactored the database logger to only register if the logger is turned on, which will be a slight performance increase.  Finally, set the log level in the core project to "info" to reduce a lot of the noise generated during the build.  If there is any issue with that, let me know.


satishc1:
@tabladrum LGTM

PR #1427, opened 2017-05-28
Make first time you start Hygieia an easier experience

virtualpeter:
If you pull Hygieia down onto a clean system, and then follow the instructions in Setup.md to build it, build docker containers, and then run docker-compose up -d it won't work, you have to add some extra config that isnt mentioned in the documentation.

There's a few small differences in the defaults with the database name (is it dashboard, or dashboarddb?)  Also the instructions for creating an empty database with user don't use the same username and password.

api doesn't want to start because the Dockerfile cmdline for java needs the jasypt config item which was added recently.


xiaods:
need rebase

virtualpeter:
seems to be ignored.    i have a better idea now anyway - so will redo and start a new pull

satishc1:
@virtualpeter @xiaods Not ignored. Just haven't gotten around to it. Thanks for the attempt at the docker cleanup. There's an older pr https://github.com/capitalone/Hygieia/pull/873 that needs rebase / merge. Will merging that cause problems for you or will you be able to rebase after it ?

xiaods:
@virtualpeter can't wait to see your pr.

Issue #1386, opened 2017-05-12
Proposal: Securing APIs


satishc1:
@tabladrum @rob-miller-777 
In terms of securing APIs with some sort of token do you think this is a valid approach?

When an external system wants to make a call to a secure API in hygieia, the hygieia admin could generate a JWT token for them. The input for creating the token would be some identifier (key) and an expiration date. We wouldn't store the JWT token in hygieia, but we would store the identifier and the expiration date to have a record that a JWT token was issued to the calling system.

The calling system would pass in the token in the header of their call. We would decode the token to see its component parts that would give us the identifier.

A JWT token isn't really meant to be revoked, but at some point if we do wish to revoke the token we could use the "identifier" to somehow block it.

rob-miller-777:
I had considered using a JWT token, but I don't think that would be appropriate in this scenario.  JWT tokens are meant to have short life cycles, since they are valid until they expire, so I was thinking that we might want to take a different approach.  My thought would be to store API keys in the database, and check them on each request.  If we start to check JWT tokens to see if they have been revoked on each request, we lose the benefit of them being stateless. We could add something to the admin page to handle generating new api keys, and managing them there.  I created #1372 , nested routes, with the idea that we could really build out the admin functionality of the app.

satishc1:
@rob-miller-777 I agree ..its better to store API keys. I can merge https://github.com/capitalone/Hygieia/pull/1372 . Do you have something already in progress for the API keys? I have some screens now but they are JWT token creation based.

rob-miller-777:
I don't have anything going for the API key stuff yet, all you.

satishc1:
@rob-miller-777 No worries. I'll get that PR merged and re-work the token screens.

rob-miller-777:
Sounds good, let me know when you have a pull request out there for the token stuff and I'll take a look.

satishc1:
Closed. We will use api keys instead of a JWT token to secure the api.

PR #1429, opened 2017-05-30
Secure APIs Basic Auth.

satishc1:
This PR provides the ability to secure APIs with an apiKey. An admin page to generate the key has been provided. The Securing APIs Basic Auth section in the api module readme provides details on how to use the apiKey.

image

image


satishc1:
@rob-miller-777 @tabladrum Please take a look and let me know your thoughts on this. Thanks.

UI/src/app/dashboard/controllers/generateApiToken.js
```diff
@@ -0,0 +1,148 @@
...
+        // Disable weekend selection
+        $scope.disabled = function(date, mode) {
+            return ( mode === 'day' && ( date.getDay() === 0 || date.getDay() === 6 ) );
+        };
+
+        $scope.toggleMin = function() {
+            $scope.minDate = $scope.minDate ? null : new Date();
+        };
+        $scope.toggleMin();
+        $scope.maxDate = new Date(2020, 5, 22);
```
rob-miller-777: Probably not a good idea to hard code the max date to the year 2020.

api/README.md
```diff
@@ -236,3 +236,27 @@ For additional information, see jasypt spring boot [documentation](https://githu
 
 Tip: If using GitLab CI Runner, specify the value for JASYPT_ENCRYPTOR_PASSWORD as a secure variable. Secure variables can be added to a Gitlab project by navigating to Project Settings > Variables > Add Variable. 
 A secure variable's value is by default not visible in the build log and can only be configured by an administrator of a project.
+
+### Securing APIs Basic Auth
+
+From the admin menu, generate an "apitoken" for an "apiuser"
+
+Create a POST request with the following 2 headers and retrieve a JWT token for secured API calls.
```
rob-miller-777: It's not actually a JWT token right?

api/src/main/java/com/capitalone/dashboard/config/WebSecurityConfig.java
```diff
@@ -99,6 +109,11 @@ protected StandardLoginRequestFilter standardLoginRequestFilter() throws Excepti
 	protected LdapLoginRequestFilter ldapLoginRequestFilter() throws Exception {
 		return new LdapLoginRequestFilter("/login/ldap", authenticationManager(), authenticationResultHandler);
 	}
+
+	@Bean
+	protected ApiTokenLoginRequestFilter apiTokenLoginRequestFilter() throws Exception {
+		return new ApiTokenLoginRequestFilter("/login/apitoken", authenticationManager(), authenticationResultHandler);
```
rob-miller-777: I don't think you are going to want to piggy back on the ldap and standard login filters.  You don't need to login with an API token, and you aren't going to want a JWT token since you already have an API Token.  You would probably want a filter that is looking at all requests for the API token, and converting that into the Authentication type your provider is looking for.

api/src/main/java/com/capitalone/dashboard/auth/DefaultAuthenticationResponseService.java
```diff
@@ -38,13 +42,35 @@ public void handle(HttpServletResponse response, Authentication authentication)
 			displayName = ((CustomUserDetails) authentication.getPrincipal()).getDisplayName();
 			emailAddress = ((CustomUserDetails) authentication.getPrincipal()).getEmailAddress();
 		}
-		Collection<? extends GrantedAuthority> authorities =
-				userInfoService.getAuthorities(authentication.getName(), firstName, middleName, lastName, displayName, emailAddress, (AuthType)authentication.getDetails());
-		UsernamePasswordAuthenticationToken authenticationWithAuthorities = new UsernamePasswordAuthenticationToken(authentication.getPrincipal(), authentication.getCredentials(), authorities);
-		authenticationWithAuthorities.setDetails(authentication.getDetails());
+
+        UsernamePasswordAuthenticationToken authenticationWithAuthorities = null;
```
rob-miller-777: This class is responsible for adding a JWT token to your response after logging in successfully, you probably don't want to use this, or modify it for the API token.

rob-miller-777:
@brennanananan and I will make a couple changes and create a pull request to your branch so you can see what we are thinking.

rob-miller-777:
@brennanananan and I have done some digging, and we think we have found a problem.  It looks like how this would work, is you would create an API key, which you would then use to login to receive a JWT.  We feel as though this wouldn't be an ideal pattern, you would probably want to utilize one or the other.  Here are the two options we are thinking.

1. API Key
Create an api key pretty much exactly how you are currently doing it.  Give the API key to the applications/jenkins plugins you want to access the api.  These applications would send the api key in the authorization header with each request to login.  Each time a request is sent, the api key would be checked to make sure it was still valid.  Would require some changes and new filters to work, and it is also stateful as the API now has to manage keys.

2. JWT
Modify the create token to create a JWT which you could give to an application.  The JWT would be good for a specific amount of time.  You could control at creation time the permission the JWT would have.  It would be very similar to authorization tokens Github uses. The benefits of this would be that it is similar to how the rest of the application already works, therefor wouldn't require many changes.  It is also stateless since it's a JWT.

UI/src/app/dashboard/controllers/generateApiToken.js
satishc1: @rob-miller-777 Will fix the date.

api/README.md
satishc1: @rob-miller-777 There are two tokens. The first one is the apitoken that the system uses to authenticate. Hygieia sends back a JWT token in response with ROLE_API authority. The calling system then uses the JWT token to make its calls.

satishc1:
@rob-miller-777 @brennanananan 

"It looks like how this would work, is you would create an API key, which you would then use to login to receive a JWT."  Correct.. that is the intended pattern. I've highlighted below the benefits of the pattern.. which you have already listed in your options.

1. API Key
**_Create an api key pretty much exactly how you are currently doing it. Give the API key to the applications/jenkins plugins you want to access the api. These applications would send the api key in the authorization header with each request to login. Each time a request is sent, the api key would be checked to make sure it was still valid._** Would require some changes and new filters to work, and it is also stateful as the API now has to manage keys.

2. JWT
Modify the create token to create a JWT which you could give to an application. The JWT would be good for a specific amount of time. You could control at creation time the permission the JWT would have. It would be very similar to authorization tokens Github uses. **_The benefits of this would be that it is similar to how the rest of the application already works, therefor wouldn't require many changes. It is also stateless since it's a JWT_**.

rob-miller-777:
How do you see this working from the jenkins plugin perspective, or anything that would want to use this pattern?  Would you give the jenkins plugin the API token, and it would login and receive a JWT each time it runs?

satishc1:
@rob-miller-777 Thats the way I envisioned it. I went back and forth on the options you had listed  when I was creating this and settled on this pattern because it seemed to have the best of both worlds.

rob-miller-777:
I'm just worried about it from a consumption perspective, I don't believe this is a common pattern.

satishc1:
@rob-miller-777 I only see benefits using this pattern because it fits how we are currently using it in Hygieia. The beauty of the JWT token is its stateless so we definitely want it in the picture. But we don't want to assign an indefinite expiration to it.

satishc1:
@rob-miller-777 Can you take this for a spin and mull it over some? I'll wait to merge this. Definitely want your opinion on this. Thanks.

rob-miller-777:
I like the idea of using JWT tokens and setting an expiration of your choosing for that token.  Could make it very similar to github in being able to assign the permissions it has as well.

satishc1:
@rob-miller-777 Yeah, but we talked about that here https://github.com/capitalone/Hygieia/issues/1386 but decided against that route.

rob-miller-777:
Right, I thought we had agreed on using API tokens instead of JWT.  The combination of the two is what I have a hard time getting on board with. From a consumption perspective, I would expect to be given a token which I could attach as a header to use as my authorization.  I don't think it's a good idea for a consuming app to need to make two calls to access a resource, when it really only needs to make one.

satishc1:
@rob-miller-777 @tabladrum Ok no worries. Let me see how big of a code change it will be and then decide whether or not to merge it and refactor later.

rob-miller-777:
Would be good to get some other opinions on this as well.

satishc1:
@rob-miller-777 @MarkRx @tabladrum Changed the apitoken auth to not require a second JWT call.  The api module readme details how to create the basic auth header for secured api calls. WebSecurityConfig will need the appropriate antMatcher with the particular api end point for example
.antMatchers("/ping").hasAuthority("ROLE_API")

Which api end points to secure could be left to the users based on some api property.

api/src/main/java/com/capitalone/dashboard/auth/apitoken/ApiTokenRequestFilter.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
+
+        HttpServletRequest request = (HttpServletRequest)req;
+        HttpServletResponse response = (HttpServletResponse)res;
+
+        String path = request.getRequestURI();
+
+        if (path.startsWith("/api/appinfo")) {
```
rob-miller-777: How come this particular endpoint is hardcoded in here?

rob-miller-777:
Currently pretty much all gets are unsecured, and all posts require authentication besides the ones for the jenkins plugins, which we made exceptions for in the web security config.  I would say we remove the exceptions once the jenkins plugin has the ability to send an api token with the request.

rob-miller-777:
I'm also a little concerned about the token itself.  It looks like you are calling it basic auth, which would be username/password.  I was thinking the header would be something like,
 `Authorization: apiToken {some random guid}`.

satishc1:
@rob-miller-777 Correct. Once the jenkin plugin handles tokens we will circle back and clear out the TODOs in WebSecurityConfig.

rob-miller-777:
Going to have to make sure to advertise that properly as it will break that interaction until people upgrade their plugin and provide an api key.

satishc1:
@rob-miller-777 Agreed. We should do a release of a new jenkins plugin first which requires a token and that can be checked during the test connection piece. We'll give it some time before we actually restrict the APIs on the hygieia side.

satishc1:
@rob-miller-777 And changing Basic to apiToken.
Authorization: Basic <authHeader>
Authorization: apiToken {some random guid}.

satishc1:
@tabladrum @rob-miller-777 LGTM

PR #1431, opened 2017-05-30
AppDynamics - Limited page size to 20

nireeshT:
AppDynamics - Limited page size to 20 for sorting timestamp


satishc1:
@tabladrum LGTM

PR #1445, opened 2017-06-01
Fix create dashboard button showing up when not logged in

rob-miller-777:
Looks like this got lost in a merge conflict.  Just hides the create button if you aren't logged in.


satishc1:
@tabladrum LGTM

PR #1446, opened 2017-06-01
Promoting Dashboard User to Dashboard Owner

brennanananan:
Currently, the modal to edit ownership of a dashboard only checks to see if the user is currently an admin of the application, and makes the user an admin if the case is no. When a user is promoted to a dashboard 'admin', they are inadvertently promoted to an admin for the entire application. For more detail, I will define what I believe to be an 'owner' vs an 'admin'. 

#### Owner
* Able to rename/delete dashboards **they own**
* Able to edit widget configuration for dashboards **they own**
* Able to edit owners list for dashboards **they own**

#### Admin
* Able to rename/delete **all** dashboards
* Able to edit widget configuration for **all** dashboards
* Able to edit owners list for **all** dashboards
* Edit list of application administrators
* Change theme
* etc

In this PR I have reduced the scope of promoting someone from an application admin, to an owner of the specific dashboard, as well as some minor cleanup (wording, method redundancy, etc).

@tabladrum @satishc1


satishc1:
@tabladrum LGTM

PR #1451, opened 2017-06-05
Fix for api losing connectivity with mongo.

satishc1:
Resolves https://github.com/capitalone/Hygieia/issues/1218


satishc1:
@tabladrum LGTM

Issue #1444, opened 2017-06-01
Rest API


rob-miller-777:
Want to take a stab at a V2 of the API. Based on your roadmap, it looks like you wanted to create a more intelligent API, here were some thoughts I had for enhancements.

- Take advantage of Spring HATEOAS, and make the API self discoverable through use of link and proper rest nouns/verbs.  Will be more consistent and easier to navigate.

- Separate the domain model from the rest api resources.  Currently, the client needs to have more knowledge of the domain than it should have to.  Since Hygieia has been around a while now, we have a better understanding of the use cases, which should give us a better understanding of what resources we will need to expose.

- Instead of removing the old API, create a V2 we could slowly migrate to.  This would most likely be accomplished through the url.  ex. "http:/localhost/api/v2/foo"

Let me know if you guys have any thoughts on this, or would like to see anything else.

mrpudn:
I'd also like to decouple the collector's from the database by having them use the API instead. This should hopefully make it a bit easier for individuals to build/maintain collectors.

rob-miller-777:
Yeah, was definitely going to keep that in mind as we go.

irvingoe:
There is some discussion on gitter regarding some changes needed to receive "Comcast" contributions like Rally collector and some others they built.

mrpudn:
@irvingoe I believe that one involves changing Hygieia from a monolithic git project to a submodule git project, which might be a bit beyond the scope of this particular issue.

We could certainly split the API out into it's own project, though - that might make it easier for individuals to set up and make localized contributions to just the API. The same could be said about the UI - if all you want to do is make a few aesthetic improvements, you (theoretically) wouldn't have to set up a ton of stuff to make that contribution.

@rob-miller-777 totally onboard with all of the above ideas, though.

tabladrum:
fyi... we are looking into having Hygieia as an org. It will help breaking things easily. Stay tuned.

tabladrum:
Long term goal is to create GraphQL api layer. Closing this.

PR #1466, opened 2017-06-09
Rest API v2

rob-miller-777:
Started to implement a v2 of the rest API.  Before we started to do the entire thing, we wanted to get a small piece out there to get some feedback on, so we rewrote the "Dashboards" controller.  We followed what we had mentioned in #1444 , and implemented v2 using Spring's hateoas library.  You will notice upon making a get request to "/api/v2/", the response will contain links to resources in the application.  Right now there are only a couple relating to dashboards, but eventually there will be more as more controllers are added.  Making a get request to the dashboards endpoint will give you the list of dashboards, and each of the dashboards will have a link to itself.  We also updated swagger to separate v1 from v2, so you can see the different versions of the api.  We tried to avoid making changes to the backend services as much as possible, but there were a couple small changes that had to be made to make this work properly, but nothing that breaks backwards compatibility.


chtompki:
This seems to have a considerable number of conflicts. Might you be able to remedy them?

rvema:
closing the PR due to inactivity

PR #1510, opened 2017-06-30
Spring upgrade

rob-miller-777:
Upgraded Spring and spring data to the latest release versions, as well as removing some unneeded dependencies.  One issue I found was since the "core" jar is technically a spring boot app, there needed to be a configuration change to the spring maven plugin in core's pom to make it work due to a packaging change from boot 1.3 -> 1.4.  We should probably refactor the poms to make core just a normal jar to avoid this.


subodhbattina:
@rob-miller-777 : Thanks for the PR, I will review it and get back to you, Thanks !

chtompki:
This pull request has a considerable number of merge conflicts, can you remediate them by chance?

tabladrum:
Lets close this... I have this in motion (within a week)

PR #1527, opened 2017-07-11
Build config modal is not retaining selected value

nireeshT:
Fix for issue- Build config modal is not retaining selected value


satishc1:
@tabladrum LGTM

PR #1433, opened 2017-05-30
Spring Cloud Config

rob-miller-777:
Added a dependency to the "core" project for spring cloud config, and made the properties builder optional by adding a skip flag.  Adding the Spring Cloud Config dependency will make all the dependent boot apps check for a cloud config server on startup.  By default it will check for a config server at "http://localhost:8888".  That can be overridden by setting the `
spring.cloud.config.uri` property.

PR #1528, opened 2017-07-12
Property builder fix

rob-miller-777:
Had to change from "return" to "exit 0" because return isn't supported how this needs to be used.  Has the same effect, but it actually works properly now.


rob-miller-777:
Just to be clear, I introduced this with #1433 .  It doesn't break anything, but won't work as intended until this is merged.

satishc1:
@tabladrum LGTM

PR #1533, opened 2017-07-14
Hygieia pubisher plugin fix

nireeshT:
Fix for Hygieia pubisher plugin not pushing build status from Jenkins to Hygieia


satishc1:
@tabladrum LGTM

PR #1538, opened 2017-07-17
docker:build - bug in api Dockerfile

virtualpeter:
pull #873  broke docker:build of api  - this change only fixes that one fault


satishc1:
@tabladrum LGTM

PR #1541, opened 2017-07-18
more docker updates

virtualpeter:
switch the docker base images from the old forgotten java to the openjdk image that they want us to use. note, the old java one was openjdk also, this change is mainly so that the java version will track security updates.

Also - switched from 8-jdk to 8-jre because that makes the images quite a lot smaller.

support for sonarqube v6 when deploying using docker.


satishc1:
@tabladrum LGTM @virtualpeter Cool. Thank you for the cleanup.

PR #1553, opened 2017-07-21
Renamed Splunk Logging Components

brennanananan:
This PR merely renames the splunk logging components to something that sounds a little more generic. While using the splunk product was the inspiration behind the components, we didn't want developers to feel like they needed splunk deployed in order to log in a key-value pair style.

@rob-miller-777


satishc1:
@tabladrum LGTM

PR #1567, opened 2017-07-25
Gitlab API v4

rob-miller-777:
Adding support for Gitlab API v4 since v3 will be removed soon.  Changes make the collector default to the v4 url, but supports v3 as well if configured properly.


satishc1:
@tabladrum LGTM

PR #1570, opened 2017-07-26
Set skipDockerBuild to true for ui-tests

satishc1:


satishc1:
@tabladrum @kumarvarun1252 LGTM

PR #1573, opened 2017-07-26
Gitlab commit type bug

rob-miller-777:
Fixed a bug I created that wouldn't allow commits to be shown on the product dashboard since all commits had at least one parent.


satishc1:
@tabladrum LGTM

PR #1576, opened 2017-07-28
Remote dashboard creation and audit api

AuditHawks:
This PR adds a framework for remote dashboard creation via api. It also adds an api-audit module which takes minimal classes from the api module and creates a framework to support audit related tasks for your pipeline.


satishc1:
@tabladrum LGTM

PR #1578, opened 2017-07-28
Fix for issue with Dashboard API 

Sbrenthughes:
Fixes Null check issue when calling Dashboard API for pulling Dashboard by Business Service/ Business Application.


satishc1:
@tabladrum LGTM

PR #1143, opened 2017-02-02
Template removal

brennanananan:
Review https://github.com/capitalone/Hygieia/issues/1116 for initial thoughts regarding this PR.
***Commits have been squashed for convenience.***

* The create dashboard modal has been expanded to allow selection of widgets from all 3 panes (1 below). 
* Update modal has been added to reselect which widgets the owner/admin wants to display (2 below).
* Configure icon has been added next to pane button set to allow for updates. This window can be expanded later to support extra dashboard specific configurations (3 below).
* Update is only available to owner/admin.
* Preexisting dashboards start with all widgets selected to support backwards compatibility, extraneous widgets can be removed by owner/admin.
* If no widgets are selected to be shown for a given pane, a helpful message appears explaining to any users wondering why no widgets are displayed.
* Template variable has been **removed** from the dashboard object, dashboard now maintains a map of available widget types and the panes where those widgets will be found.
* All template views have been consolidated into one single 'team.html/js' view. 

**Note** This PR allows users to select which widgets are displayed, it does not allow for multiple widgets of the same type. IE, having 2 Repository widgets is not supported.

**Example Screens**

1
screen shot 2017-02-02 at 1 16 11 pm

2
screen shot 2017-02-02 at 1 17 31 pm

3
screen shot 2017-02-02 at 1 22 40 pm

PR #1581, opened 2017-07-31
Custom template creation for team dashboard

nireeshT:
This PR includes below functionality
1) Create custom templates for team dashboards as admin
2) View/Edit/Delete custom templates


tabladrum:
@nireeshT Can you throw in some screen shots?

nireeshT:
1) Login as admin
2) Go to settings -> Manage Templates -> Create Template 
screen shot 2017-07-31 at 4 22 08 pm
screen shot 2017-07-31 at 4 18 55 pm
screen shot 2017-07-31 at 4 19 16 pm
Note : Order which you have selected the widgets on the layout is the same order dashboard will have widgets aligned.
3) Edit template : 
screen shot 2017-07-31 at 4 18 05 pm
4) Edit template confirmation screen 
screen shot 2017-07-31 at 10 47 37 am
5) Delete Template : Template can't be deleted if a dashboard is configured. Dashboard should be deleted prior to template deletion. 
screen shot 2017-07-31 at 10 46 43 am
6) Dashboard creation : Select created template in dropdown in create dashboard modal
screen shot 2017-07-31 at 4 19 37 pm
7) Dashboard will be aligned with the widgets with equal dimensions (3 widgets in a row) 
screen shot 2017-07-31 at 4 20 17 pm
8) Editing template (insertions and deletions of widgets) after creating dashboard will adjust the existing dashboards widget order and place the new added widgets at the end.

rob-miller-777:
Will this give us the ability to remove the existing templates, such as "Cap One", and "Split View", or just add new ones to the list?

nireeshT:
@rob-miller-777  - It will add new ones to the list keeping the existing ones intact.

rob-miller-777:
I don't know that I see the value in having templates for team dashboards.  In my opinion, first thing you should do is choose your dashboard type, (Team, Product, executive in future), name your dashboard, then click create.  If you choose a team dashboard, your page will initially be empty, and at that point you would configure the widgets you want to see, similar to your manage templates screenshot above.  Then you can go about configuring each of the widgets on your page. I also think it would be a lot more valuable for users to have the ability to add/remove widgets even after creating their dashboards, which could be accomplished with this flow.

brennanananan:
I agree with Rob. The original intention behind #1143 was to allow the users to see which ever widgets they wanted. Having templates managed by a single (admin) source still restrains the consumer to whatever templates exist. I do really like the way that manage template screen turned out, but if that could be a "Widget Selection" screen or something per dashboard creation that would fit the intention of my PR a little more.

nireeshT:
@rob-miller-777 @brennanananan  - I do agree with you where users can add/remove the widgets on their respective dashboards. But the main intent of this PR is to adhere reusability and maintainability. Manage templates enables to create different templates and those can be reused across and always there is an option to edit as required. Through enterprise level perspective, we may have hundreds of dashboards which may become hard to manage when they are configured at dashboard level with enterprise standards(enforce to use only selected widgets). So we did move manage template functionality under admin which enables single source for configuration. We can have hybrid model where a custom template is created, give access to user to hide/show a widget and set a flag(enabled/disabled) to see widgets only they wanted. This change will be persistent and once they come back to dashboard, they see all the dashboards enabled to see. In this way, control will be at admin(for creating custom templates) and with the user to enable/disable the view of widgets at dashboard level.

Use case
1) Admin creates custom template with widgets -
 Build,Deply,Feature,Performance,Quality,CodeRepo
2) User uses custom template to create a team dashboard. 
3) On dashboard we will enable configure button which will open a dialog box which user can select/deselect the existing widgets on dashboard.

rob-miller-777:
So is the functionality you are looking for the ability to control what widgets are available for users at an admin level?

nireeshT:
@rob-miller-777  - Build , CodeRepo, Feature, Monitor, Deploy, Performance, Quality. It can be extended to pipeline and Cloud as well.

rob-miller-777:
I'm asking from a higher level.  For example, Company A doesn't want their employees to use the monitor widget, so they create a template without it.  Is that your intended use case?

nireeshT:
@rob-miller-777  exactly

rob-miller-777:
For that use case, I think I would rather see an admin page for toggling available widgets rather than controlling it through templates.  Especially if the default templates are going to stick around, because if you wanted to get around the custom templates not having the monitor widget, you could just create a "cap one" dashboard.

nireeshT:
@rob-miller-777  means admin-> Edit Dashboards functionality should be extended to configure widgets as well? like edit dashboard title, owner information, we should have one more tab to choose the widgets for dashboard. correct me if I am wrong.

rob-miller-777:
No, I am thinking we have a tab on the admin page where you can disable a widget application wide.  So you couldn't add a certain widget type to any dashboard in the app if you didn't want to.

nireeshT:
@rob-miller-777  - Sounds good, what if we want to disable different widgets for different applications with in enterprise.

rob-miller-777:
Should probably work similar to the feature collector widget where it doesn't even show up as an option until the collector has run.

satishc1:
@rob-miller-777 @brennanananan I think what @nireeshT is saying is that by having templates, you can choose to disable different widgets for different user bases.

If we go the route of adding widgets directly from the admin tab, I don't see how you could control disabling a widget for a certain user but enabling that same widget for another user.

rob-miller-777:
I would be fine to keep templates if they weren't mandatory for the user. So create a "Team Dashboard".  Then on the team dashboard page, you could add the widgets you want, or choose a template.  And even if you choose a template, it would just be a starting point, so you could add or remove afterwards.

brennanananan:
I agree with Rob. Templates should be optional, but a good starting point if the user chooses so.

satishc1:
@rob-miller-777 @brennanananan That makes sense. @nireeshT Lets discuss.

nireeshT:
@rob-miller-777 @brennanananan  
We have discussed the design and this how we plan to implement.
1. On the Create Team Dashboard dialog box, we’ll have a toggle where the users can choose either Widgets or Templates. If they choose widgets, it will prompt them with the re-worked Template dialog box but its just going to create widgets directly.
If they choose templates, then it supports the existing flow.
2. Templates need to be supported because existing dashboards need them. You could choose to delete all dashboards and recreate them, but that would be a lot of work.
3. On the existing “Configure your Dashboard” dialog box, we’ll add a tab "configure widgets" that supports re-configuring your widgets if you didn’t choose a template to begin with.
4. The Admin screen will have a template option.
Let me know your thoughts.

brennanananan:
@nireeshT I disagree with your second point, my solution was to just show all widgets for "legacy" dashboards until users decide to hide elements they don't want. Otherwise I am fine with what you've proposed.

nireeshT:
@tabladrum @satishc1 @rob-miller-777 @brennanananan  - Updates for PR includes 
1) Enable option to create dashboard by selecting widgets or selecting pre-existing/custom templates
2) Widget Management modal appears when user choose to select widgets option and will be able to save selected widgets to create dashboard.
3) Existing dashboard widgets can be edited (Add/Remove widgets) via -> edit dashboard-> Widget Management
screen shot 2017-08-22 at 1 03 47 am
screen shot 2017-08-22 at 1 04 00 am
screen shot 2017-08-22 at 1 04 27 am
screen shot 2017-08-22 at 1 04 41 am
screen shot 2017-08-22 at 1 04 53 am
screen shot 2017-08-22 at 1 05 23 am

MarkRx:
I like the idea of being able to customize the widgets that show up in a dashboard. I can see cases where people want to exclude certain widgets or replace them with others. This may become more useful as more widgets are developed.

As for individual dashboards I can see it being useful to have a small set of global templates but allowing the owners of each dashboard to further customize it as desired.

rob-miller-777:
Definitely like that better. Good stuff.

brennanananan:
LGTM - closing #1143 in response

satishc1:
@tabladrum LGTM

PR #1594, opened 2017-08-07
Appdynamics docker

virtualpeter:
calling docker:build on the whole project fails because appdynamics-collector does not have files to support this method and hygieia-jenkins-collector does not have handler for maven-docker.


virtualpeter:
i did a tidier version of this pull request

PR #1595, opened 2017-08-07
Changes for deletion confirmation dialog box.

NeetiM:
Changes done for delete dashboard confirmation


nireeshT:
@NeetiM  - can you throw in some screenshots

nireeshT:
@NeetiM  - We may want confirmation alert here instead of a modal popup to ask for delete. You can use swal(SweetAlert.js) to set alerts. I would create a custom directive to show confirmation alert and use it across application on delete button element (or any element which would require confirmation alert).

NeetiM:
image

NeetiM:
image
image
image

Issue #1568, opened 2017-07-25
Delete dashboard should ask for confirmation


kumarvarun1252:
Right now deleting a dashboard does not ask for a confirmation. It would be really helpful if there is a confirmation for delete dashboard. I accidentally clicked on the delete button and I couldn't get the dashboard back.

nitin1706:
Good point, will have a look and put it to fix.

nitin1706:
PR #1608

nitin1706:
@Sbrenthughes : can you close this issue

nitin1706:
@Sbrenthughes :  we can close this issue

PR #1608, opened 2017-08-09
Changes for custom directive for deletion confirmation

NeetiM:
This directive "deleteConfirm" can be used for the deletion confirm for any component across Hygieia.
For this, we just need to add attributes:
1. "delete-confirm" - that invokes a sweet alert for confirmation with cancel and delete buttons
2.  confirm-action -the function called for deletion, needs to be assigned to this attribute .

For example :
<a href id="deleteDashboardButton" delete-confirm confirm-action="ctrl.deleteDashboard(myitem)">
                                          <span class="fa fa-lg fa-trash text-danger"></span>
       </a>


nitin1706:
@satishc1 : we can close this PR as I have created a new one with complete changes

nitin1706:
PR for #1568

satishc1:
@tabladrum LGTM

Issue #1612, opened 2017-08-10
Exception while fetching feature estimate for Jira collector


nitin1706:
I am seeing multiple NumberFormatException when Hygieia is trying to fetch feature estimate for Jira Collector.

PR #1613, opened 2017-08-10
added null check for sonar properties and exception handler for jira …

nitin1706:
…feature values


nitin1706:
Fix for issue #1612

satishc1:
@tabladrum LGTM

PR #1614, opened 2017-08-10
fixed clound typo

mateusz-kula:
There was a minor typo in AWS cloud client "cloud" was mistyped as "clound", the only change is to remove "n"


satishc1:
@tabladrum LGTM

PR #1620, opened 2017-08-14
Fix apiaudit target, uib-typeahead for quality widget, null ptrs for jenkins job config history

satishc1:


satishc1:
@tabladrum @nireeshT LGTM

Issue #1633, opened 2017-08-17
Jenkins-plugin: wrong concatenation of workspace and Artifact Directory in Publish Deployment Results


schup:
The Artifact directory currently requires a leading / otherwise artifacts are not found correctly.

A leading / for a workspace relative path is not intuitive.

Spotted this in the logs - no / between workspace and build
`Hygieia Deployment Publisher - Looking for file pattern '*.jar' in directory /var/lib/jenkins/jobs/com/jobs/service-b/branches/master/workspacebuild/libs`

PR #1634, opened 2017-08-17
#1633 handle missing leading slash in artifact directory

schup:
By using the FilePath constructor leading slashes or absence of it is handled transparently.

Resolves https://github.com/capitalone/Hygieia/issues/1633


satishc1:
@tabladrum LGTM

Issue #1257, opened 2017-03-22
Jenkins collector StackOverflowError


MarkRx:
Either the mongo save listener got stuck in a loop or the stack is huge. Stack trace below:

> 2017-03-22 08:51:10,879 DEBUG c.c.d.collector.DefaultHudsonClient -  Adding Build: http://jenkins.com:8080/job/job/1/
> 2017-03-22 08:51:10,879 INFO  c.c.d.collector.CollectorTask - Fetched jobs                    10s
> 2017-03-22 08:51:11,038 INFO  c.c.d.collector.CollectorTask - New jobs                0        0s
> 2017-03-22 08:51:11,054 DEBUG c.c.d.collector.DefaultHudsonClient - Enter makeRestCall http://jenkins.com:8080/job/myjob/job/master/9/api/json?tree=number,url,timestamp,duration,building,result,culprits[fullName],changeSets[items[user,author[fullName],revision,id,msg,timestamp,date,paths[file]],kind],changeSet[items[user,author[fullName],revision,id,msg,timestamp,date,paths[file]],kind,revisions[module,revision]],actions[lastBuiltRevision[SHA1,branch[SHA1,name]],remoteUrls]
> 2017-03-22 08:51:11,149 DEBUG c.c.d.collector.DefaultHudsonClient - Enter makeRestCall http://jenkins.com:8080/job/myjob/job/master/9/consoleText
> 2017-03-22 08:51:18,944 ERROR o.s.s.s.TaskUtils$LoggingErrorHandler - Unexpected error occurred in scheduled task.
> java.lang.StackOverflowError: null
> 	at java.nio.charset.CharsetDecoder.replaceWith(Unknown Source) ~[na:1.8.0_121]
> 	at java.nio.charset.CharsetDecoder.<init>(Unknown Source) ~[na:1.8.0_121]
> 	at java.nio.charset.CharsetDecoder.<init>(Unknown Source) ~[na:1.8.0_121]
> 	at sun.nio.cs.UTF_8$Decoder.<init>(Unknown Source) ~[na:1.8.0_121]
> 	at sun.nio.cs.UTF_8$Decoder.<init>(Unknown Source) ~[na:1.8.0_121]
> 	at sun.nio.cs.UTF_8.newDecoder(Unknown Source) ~[na:1.8.0_121]
> 	at java.lang.StringCoding.decode(Unknown Source) ~[na:1.8.0_121]
[...]

PR #1638, opened 2017-08-17
added a null check- fix 

nitin1706:
reworked


nitin1706:
#1257

satishc1:
@tabladrum LGTM

PR #1651, opened 2017-08-22
changes for password encryption issue

nitin1706:
Complete changes for custom directive for deletion confirmation and password encryption.


satishc1:
@tabladrum LGTM

PR #1656, opened 2017-08-24
Update setup.md

mfarouk:
Changing the database setup from "use dashboarddb" to "use dashboard" and the output message from "switched to db dashboarddb" to "switched to dashboard". The db name is dashboard so this setup causes issues when setting up other properties files and calling the database dashboard and the dashboard user has no readWrite permissions to it.

### DIFF
--- a/pages/hygieia/setup.md
+++ b/pages/hygieia/setup.md
@@ -35,8 +35,8 @@ The following components are required to run Hygieia℠:
          MongoDB shell version: 3.0.4
          connecting to: test  
          
-         > use dashboarddb
-         switched to db dashboarddb
+         > use dashboard
+         switched to db dashboard
          > db.createUser(
                   {
                     user: "dashboarduser",


satishc1:
@Megha849 Please take a look and let me know.

MeghaMyWork:
@mfarouk We are in the process of re-organizing our documents. For the suggested change, please refer to the following link:
http://www.capitalone.io/Hygieia/database.html
The setup.md document will be deleted once the re-organization is complete.

mfarouk:
@MeghaMyWork that looks nice! However, in the new doc you're calling the db dashboarddb while it is called dashboard almost everywhere else and in all the properties files examples, even in your example under http://www.capitalone.io/Hygieia/api.html
I suggest that you call it dashboard instead to make the setup process simpler. Also doesn't hurt to fix it in setup.md while the new docs are being written.

satishc1:
@mfarouk @MeghaMyWork dashboarddb is what needs to be in the documentation. At one point it used to be dashboard.

PR #1658, opened 2017-08-24
Update mongosrc.js

mfarouk:
Changing the database setup from "use dashboarddb" to "use dashboard". The db name is dashboard so this setup causes issues when setting up other properties files and calling the database dashboard and dashboard user has no readWrite permissions to it. Similar to change proposed in pull request #1656 to setup.md.

### DIFF
--- a/mongosrc.js
+++ b/mongosrc.js
@@ -1,5 +1,5 @@
 use
-dashboarddb
+dashboard
 db.createUser(
     {
         user: "dashboarduser",


satishc1:
@Megha849 Please take a look and let me know.

MeghaMyWork:
@mfarouk We are in the process of re-organizing our documents. For the suggested change, please refer to the following link:
http://www.capitalone.io/Hygieia/database.html
The setup.md document will be deleted once the re-organization is complete.

mfarouk:
This proposed change is for the setup script (mongsrc.js file) and not for for the setup markdown. My comment only says it's the same change as what I am proposing in setup.md.

satishc1:
@mfarouk @MeghaMyWork dashboarddb is what needs to be in the documentation. At one point it used to be dashboard.

PR #1663, opened 2017-08-24
Update setup.md

mfarouk:
@satishc1 @Megha849 Changing to dashboarddb. I'll try to make the change in the other properties files as well. I think this will minimize confusion and questions. Thanks!


satishc1:
@tabladrum LGTM

PR #1674, opened 2017-08-27
add dockerfile and parameter builder for appdynamics collector

virtualpeter:


MarkRx:
Please get this merged :)

MarkRx:
Changes included with #1708

collectors/performance/appdynamics/docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM docker.io/openjdk:8-jre
+
+MAINTAINER Hygieia@capitalone.com
+
+
+ENV SPRING_DATA_MONGODB_DATABASE=dashboard
+ENV SPRING_DATA_MONGODB_HOST=10.0.1.1
+ENV SPRING_DATA_MONGODB_PORT=9999
+ENV SPRING_DATA_MONGODB_USERNAME=db
+ENV SPRING_DATA_MONGODB_PASSWORD=ENC(aSPTk36yA/ZklUg75CrZ8w==)
```
satishc1: @virtualpeter Please take these out. The default values are set in the prop builder.

virtualpeter: these lines are necessary to support docker-compose v2 and help when debugging this declares the variables so they can be passed in at runtime -  i followed the pattern set by all the other collectors' Dockerfile and prop builders in that they are defaulted here as well as in the collector.

asg247: @virtualpeter the ENV variables can be passed through docker-compose.yml file.

collectors/performance/appdynamics/docker/Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+  mkdir /hygieia
+
+COPY hygieia/ /hygieia
+COPY appdynamics-properties-builder.sh /hygieia/
+
+WORKDIR /hygieia
+
+VOLUME ["/hygieia/logs"]
+
+EXPOSE 8080
```
asg247: Not required to expose any port for collectors

asg247:
Couple of issues with this PR.

1. The appdynamics-properties-builder.sh does not has execute permission, it should be set as 755. Otherwise, it wont be executed when container starts.
2. logback.xml is missing.

I can submit a PR on top of this PR to fix all the issues.

satishc1:
@tabladrum LGTM

PR #1676, opened 2017-08-28
Gitlab Feature Subgroup Support

rob-miller-777:
The big change here is moving "teamId" from a path param to a query param.  Gitlab has a concept of subgroups, and the subgroups are delimited by "/".  You can't send a "/" as a path param obviously, and url encoding it is still not accepted by most servers unless explicitly turned on.  The other change was moving the query parameters in the angular app into an object, instead of doing string manipulation which is very brittle.


nireeshT:
@satishc1  @tabladrum  LGTM

satishc1:
@tabladrum LGTM

Issue #1317, opened 2017-04-13
Null ptr when getting dashboard.


satishc1:
```
2017-04-13 11:53:42,189 ERROR c.c.d.rest.RestApiExceptionHandler - null
java.lang.NullPointerException: null
	at com.capitalone.dashboard.service.DashboardServiceImpl.get(DashboardServiceImpl.java:66) ~[api.jar!/:2.0.5-SNAPSHOT]
	at com.capitalone.dashboard.rest.DashboardController.getDashboard(DashboardController.java:70) ~[api.jar!/:2.0.5-SNAPSHOT]
	at sun.reflect.GeneratedMethodAccessor461.invoke(Unknown Source) ~[na:na]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_66]
	at java.lang.reflect.Method.invoke(Method.java:497) ~[na:1.8.0_66]
	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:221) ~[spring-web-4.1.7.RELEASE.jar!/:4.1.7.RELEASE]
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:137) ~[spring-web-4.1.7.RELEASE.jar!/:4.1.7.RELEASE]
	at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:110) ~[spring-webmvc-4.1.7.RELEASE.jar!/:4.1.7.RELEASE]
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandleMethod(RequestMappingHandlerAdapter.java:776) ~[spring-webmvc-4.1.7.RELEASE.jar!/:4.1.7.RELEASE]
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.handleInternal(RequestMappingHandlerAdapter.java:705) ~[spring-webmvc-4.1.7.RELEASE.jar!/:4.1.7.RELEASE]
	at org.springframework.web.servlet.mvc.method.AbstractHandlerMethodAdapter.handle(AbstractHandlerMethodAdapter.java:85) ~[spring-webmvc-4.1.7.RELEASE.jar!/:4.1.7.RELEASE]
[...]

PR #1688, opened 2017-08-29
fix for Null when loading dashboard

nitin1706:
fix for #1317


nitin1706:
@nireeshT

nireeshT:
@nitin1706  - what is the use case? can you reproduce this issue ?

nitin1706:
fix for issue #1317, 
Case 1: Re-produced this issue(#1317) by deleting components and then try fetching dashboards. 
Case 2: For delete scenarios, in case, it need to delete component first and then delete dashboard. Use Case: in case, components are not deleted, it will fail with error message that dashboard cannot be deleted, while dashboard actually would have been deleted.

satishc1:
Closed.

PR #1691, opened 2017-08-30
Fix incorrect docker image names #1690

### LINKED ISSUE #1690: Bamboo and Jenkins codequality collector docker configs incorrect
Both are using the name "hygieia-jenkins-build-collector". This causes the real hygieia jenkins build collector docker image to get overwritten.

MarkRx:
Both the bamboo and jenkins codequality collectors are using the docker image name "hygieia-jenkins-build-collector". This causes the real hygieia jenkins build collector docker image to get overwritten.

Resolves #1690


satishc1:
@tabladrum LGTM

PR #1695, opened 2017-08-31
Fix +x permissions on docker properties builders that were missing them

MarkRx:
The artifactory and xldeploy properties builders were missing the +x file permission in git. This was resulting in a permission denied when trying to run them in docker.


satishc1:
@tabladrum LGTM

PR #1696, opened 2017-08-31
Fix accept header for xml to xld collector

MarkRx:
XLD 6.x seems to default to returning json on a query where it used to default to returning XML. Add an accept header to make sure we get XML back.


MarkRx:
Random npm build failure with missmatching checksum. Pushed a rebased version to trigger a new build.

satishc1:
@tabladrum LGTM

PR #1699, opened 2017-08-31
correct issue with job container not having default constructor

stevegal:
fixes JobContainer conversion from json with performing get on jenkins job request


satishc1:
@tabladrum LGTM

Issue #1207, opened 2017-02-24
API docker-compose defaults are not properly set


MarkRx:
The current state of the docker-compose file results in exceptions when attempting to run API. See below. The docker-compose API should not default to requiring Jasypt. Right now it does because the docker image is build with "ENV SPRING_DATA_MONGODB_PASSWORD=ENC(aSPTk36yA/ZklUg75CrZ8w==)" which I presume is "password" encrypted with a key of "hygieiasecret".

The following should be done:
- There should be a way to run without Jasypt. Right now I can't start my collectors with docker-compose (without overriding commands) because the docker images do not have jasypt.encryptor.password. It seems this property is required even if I am not using ENC. It could be defaulted to "unencrypted" in the code perhaps.
- The 5 ENV commands in the Dockerfile should be removed. None of the other Dockerfiles contain these.
- ~~All of the other collectors should support automatically setting -Djasypt.encryptor.password if set using the same mechanism as the API.~~ The JASYPT_ENCRYPTOR_SECRET variable should be removed along with -Djasypt.encryptor.password=$JASYPT_ENCRYPTOR_SECRET. Spring will automatically set jasypt.encryptor.password if JASYPT_ENCRYPTOR_PASSWORD is set.
- Starting the jira collector requires the JASYPT_ENCRYPTOR_PASSWORD (equal to jasypt.encryptor.password) to be set to "hygieiasecret" even through I am not using ENC. Why?

```
hygieia-api:
  image: hygieia-api:latest
  container_name: hygieia-api
  ports:
  - "8080:8080"
  volumes:
  - ./logs:/hygieia/logs
  environment:
  - MONGO_PORT=tcp://10.0.2.2:27017
  - SPRING_DATA_MONGODB_DATABASE=dashboard
  - FEATURE_DYNAMIC_PIPELINE=enabled
```

> hygieia-api                               | Caused by: java.lang.IllegalArgumentException: Password cannot be set empty
> hygieia-api                               |     at org.jasypt.commons.CommonUtils.validateIsTrue(CommonUtils.java:150)
> hygieia-api                               |     at org.jasypt.encryption.pbe.StandardPBEByteEncryptor.resolveConfigurationPassword(StandardPBEByteEncryptor.java:792)
> hygieia-api                               |     at org.jasypt.encryption.pbe.StandardPBEByteEncryptor.cloneAndInitializeEncryptor(StandardPBEByteEncryptor.java:486)
[...]

PR #1708, opened 2017-09-05
Move docker config file location to allow for mounting configs

MarkRx:
A blind mount allows you to overwrite the contents of an existing directory in docker. This PR moves the properties files for each docker image to a separate /config folder. This allows the config folder to be mounted to use config files on the local file system by setting SKIP_PROPERTIES_BUILDER=true. Doing so is advantageous as it is more flexible and powerful and because the docker properties builders often are overlooked when new properties are added.

I also fixed a part of #1207 that was missed - the ENC() block should not be in the default docker images. Since it was in hygieia-api it requires all images to use jasypt with no way to disable it. To use jasypt nothing changes - pass in / store in a compose file a JASYPT_ENCRYPTOR_PASSWORD and the property encrypted with ENC().

This also includes changes from #1674 as I needed to update that as well for this.

Docker blind mounts: https://docs.docker.com/engine/admin/volumes/bind-mounts/

Example:
```
  hygieia-api:
    image: hygieia-api:latest
    container_name: hygieia-api
    ports:
    - "8080:8080"
    volumes:
    - ./logs:/hygieia/logs
    - ./config:/hygieia/config
    environment:
    - MONGO_PORT=tcp://10.0.2.2:27017
    - SKIP_PROPERTIES_BUILDER=true
```

```
$ ls config
hygieia-api.properties                                
hygieia-appdynamics-performance-collector.properties  
hygieia-artifactory-artifact-collector.properties     
hygieia-bitbucket-scm-collector.properties            
hygieia-jenkins-build-collector.properties
hygieia-jira-feature-collector.properties
hygieia-sonar-codequality-collector.properties
hygieia-xldeploy-deployment-collector.properties
```

You can also achieve the same using docker run passing in environment variables and a mount.


matthewbeasley:
@MarkRx, I'm currently reviewing this PR and I have a couple of concerns. First off, I think it's a really good idea to pass in a properties file instead of using env vars to create a properties file at run-time. That being said, I've ran into some issues.
- 1: I'm currently using Docker For Mac version 17.09.0-ce and have mounted my config folder as you have suggested in your example, but unfortunately it seems that the folder is mounted too late in the container creation process and therefore isn't present when the last line of the Dockerfile is executed. So with a compose entry like this:
`hygieia-api:
  image: hygieia-api:latest
  volumes:
  - ./logs:/hygieia/logs
  - ./config:/hygieia/config
  environment:
    - SKIP_PROPERTIES_BUILDER=true`
and running `docker-compose up` no config is loaded since the properties builder doesn't run and the volume isn't mounted yet. As far as I can tell the only way we can get the properties file loaded in time is by modifying the Dockerfile to COPY the file over.
- 2: This issue isn't as much of a deal breaker, but still a problem. You used $MONGO_HOST and $MONGO_PORT. These are only available when using [legacy container linking](https://docs.docker.com/engine/userguide/networking/default_network/dockerlinks/) which is a problem in that it we will have to take it out later when we introduce networks and we **will** have to introduce networks at some point. Especially if we ever want to run our containers in production. So changing the names of these env vars really isn't that big of a deal since there isn't really a standard but we shouldn't assume that Mongo is always going to provide these for us.

Here's an example of the env sharing present when using linking but missing using a network:
`# save this as links.yml
# run 'docker-compose -f links.yml up | grep alpine'
# env vars are shared when using legacy container linking
mongo:
  image: mongo

alpine:
  image: alpine
  command: env
  links:
    - mongo:mongo`

`# save this as network.yml
# run 'docker-compose -f network.yml up | grep alpine'
# with networks there is no env var sharing
version: "3"

services:
  mongo: 
    image: mongo
    networks:
      - my-network

  alpine:
    image: alpine
    command: env
    networks:
      - my-network

networks:
  my-network:`

satishc1:
Thank you @MarkRx @matthewbeasley

satishc1:
@tabladrum LGTM

PR #1718, opened 2017-09-08
Defect/missing options

stevegal:
not sure why, but it would appear that the api now required some extra options in order to not null pointer. Could't find what they should be, so I've set them to what seems reasonable.


satishc1:
@tabladrum LGTM

Issue #1517, opened 2017-07-05
Sonar version 6.1 does not support /api/project_analyses/ but collector calls for it


LukeOfWales:
Collector is configured to pull projects from a Sonar server running v6.1 - so this was specified in the properties file - also using the correct sonar.metrics line (for 5.4 & above).
The following was output when this was used:
![image](https://user-images.githubusercontent.com/6364902/27860954-b363b31a-6176-11e7-8238-f993ca8ac813.png)
Throwing a 404 looking for: 
http://{SonarServer}:9000/sonarqube/api/project_analyses/search?project={ProjectName}

After researching, I found that project_analyses is not used unti v6.3 - https://next.sonarqube.com/sonarqube/web_api/api/project_analyses

The version of Sonar I am using is 6.1 however, when I change to <6 in the properties file - the collector works as expected: 
![image](https://user-images.githubusercontent.com/6364902/27861130-65d06ab6-6177-11e7-9c74-237887f78855.png)

Is it possible to investigate this issue? 
As currently I am using what appears to be a pre v6 configuration on v6.1.
```
Hygieia version: 2.0.5
Sonar version: 6.1
```

Thanks

pteichner:
```
Hygieia version: 2.0.5
Sonar version: 6.1
```

asg247:
Tracking this with related issue #1704.
Closing this Issue.

Issue #1311, opened 2017-04-11
Sonar collector is not working with Sonarqube 6.3


tinabear:
"api/resources" is removed from Sonarqube 6.3. Right now, our Hygieia sonar collector is complaining 404 error. Is there anything I can do to fix this issue?

Thank you.

rsans:
try to modify DefaultSonarClient.java with this values:
 private static final String URL_RESOURCES = "/api/components/search?qualifiers=TRK&ps=10000";
    private static final String URL_RESOURCE_DETAILS = "/api/measures/component?componentId=%s&metricKeys=%s";

Then debug methods getProjects and currentCodeQuality to fix parser errors.

tinabear:
I tried with URL_RESOURCES = "/api/projects/index" since getProjects needs a Json array. But for URL_RESOURCE_DETAILS, /api/measures/component does not return a Json array. So the collector is still not working.

rsans:
Sonar 6.3 has changed api response, so you need to change more code. Try to debug methods getProjects and currentCodeQuality and you will see what you need to modify.

tinabear:
6.3 /api/measures/component does not return some values such as "alert_text" "frmt_val" and "version". I am not sure if i can skip these values for CodeQuality and CodeQualityMetric.

rsans:
You can add metrics to response. Just see available metrics at https://docs.sonarqube.org/display/SONAR/Metric+Definitions
and add what you need to sonar collector properties file.

tinabear:
I don't worry about what metrics i can add. Just concern what data 6.3 API can return.

rsans:
....

application.properties file -> sonar.metrics properties -> add this values: alert_status,quality_gate_details

Restart hygieia sonar collector. Sonarqube 6.3 api response now returns alert_status and quality_gate_details metrics

![image](https://cloud.githubusercontent.com/assets/1969636/25217558/ab3ad0fc-25a7-11e7-8a27-65f119eebce6.png)

I've made the changes I have told you in my code and it works.

![image](https://cloud.githubusercontent.com/assets/1969636/25217685/2052eb2c-25a8-11e7-942c-a2286d71563e.png)

The correct way is to add a new sonar client in the collector code, if I have time I will create a new branch and a pull-request. Right now I made the changes directly to the client by default.

tinabear:
Awesome. But we cannot get the version from 6.3 API response, can we?
[...]

PR #1736, opened 2017-09-22
Fix for sonar collector for v6.0-v6.2

asg247:
Fix for issue #1517 #1311  Sonarqube v6.0-v6.2 returning 404 
Changes : 
DefaultSonar6Client has sonar API URLs which  is supported only for sonar 6.3 and above. 
For sonar version 6.2 and below DefaultSonarClient should be used.


satishc1:
@tabladrum LGTM

PR #1748, opened 2017-09-28
Add statuses for dashboard not registered and no pull reqs for a date range.

satishc1:


satishc1:
@tabladrum @PanchoPuncho LGTM

PR #1749, opened 2017-09-28
Format Collectors

Megha849:
@satishc1 : 
- Aligned content for all collectors in the approved format.


satishc1:
@tabladrum LGTM

PR #1770, opened 2017-10-11
Approval management

### LINKED ISSUE #1735: Audit service improvements
I think there are some improvements that can be made to the audit service. It looks like the service scans through the pull request comments and the pull request review comments. If any comments are present (and not by the PR author) then the pull request is considered reviewed.

See https://github.com/capitalone/Hygieia/blob/200f9154a9956283eb16ce01b4c5aa16b43b4d5a/api-audit/src/main/java/com/capitalone/dashboard/service/AuditServiceImpl.java#L290-L311

There are a couple of improvements that can be made:

A pull request review has a state. I see the states "PENDING", "APPROVED", "DISMISSED". There should also be a "Request Changes" status but I don't know is the corresponding API state. I think the service should be checking the state of the pull request review instead of looking at the review comments. The current implementation will accept "Pending" or "Request Changes" pull request reviews. That seems wrong. Here's the GitHub API: https://developer.github.com/v3/pulls/reviews/#get-a-single-review

There are external services that provide approval management. For example, Hygieia uses https://github.com/lgtmco/lgtm. It would be useful to allow a repository to configure Hygieia to recognize these services for approval management. These services work by setting a git commit status. The repository could register the status context with Hygieia, such as "approvals/lgtm" and that status check could be tested. https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref

Currently the audit service tests for the presence of any pull request comments. This doesn't seem like the correct behavior. For example, I could make a comment "Don't merge this. There's a NullPointerException on line 10" and that would be considered as peer review. I would remove the testing of pull request comments. Another option is to reproduce the functionality of a service like https://github.com/lgtmco/lgtm. It's a non-trivial implementation to reproduce that functionality.

I'm happy to submit pull requests on any of the three comments above. Just wanted to get feedback before I start writing code.

mspiegel:
Tests peer review using either GitHub commit statuses or GitHub reviews. If the Hygieia API settings are populated with a list of peer review contexts then test the commit statuses.  When a commit status with a matching context is found, then use the state of the commit status to indicate peer review. If the Hygieia API settings do not have a peer review context then do not use the commit status.

When no peer review commit status is found then fallback to using GitHub Reviews. Check for at least one approved review. GitHub Reviews does not allow self-review. Closes #1735

### DIFF
--- a/api-audit/src/main/java/com/capitalone/dashboard/service/AuditServiceImpl.java
+++ b/api-audit/src/main/java/com/capitalone/dashboard/service/AuditServiceImpl.java
@@ -17,6 +17,7 @@ import com.capitalone.dashboard.model.Dashboard;
 import com.capitalone.dashboard.model.GitRequest;
 import com.capitalone.dashboard.model.JobCollectorItem;
 import com.capitalone.dashboard.model.RepoBranch;
+import com.capitalone.dashboard.model.Review;
 import com.capitalone.dashboard.model.Widget;
 import com.capitalone.dashboard.repository.BuildRepository;
 import com.capitalone.dashboard.repository.CmdbRepository;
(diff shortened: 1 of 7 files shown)


rbtr:
👍

satishc1:
@PanchoPuncho Improvements based on community feedback.

api-audit/src/main/java/com/capitalone/dashboard/service/AuditServiceImpl.java
```diff
@@ -339,21 +372,7 @@ private void getRelatedCommits(String aCommit, HashMap<String, Commit> mapCommit
...
-                }
-            }
-            List<Comment> reviewComments = pr.getReviewComments();
-            for(Comment comment: reviewComments) {
-                if (!comment.getUser().equalsIgnoreCase(prAuthor)) {
-                    peerReviewed = true;
-                    break;
-                }
-            }
+            boolean peerReviewed = computePeerReviewStatus(pr);
```
satishc1: @mspiegel The check for the comments and reviewComments needs to be added as a fallback if the above compute returns false.

satishc1:
@mspiegel @tabladrum merging this as discussed and will make another PR on top of this one to address the fallback case.

satishc1:
LGTM

PR #1771, opened 2017-10-11
Added Numeric Check to story points value

irvingwa:
From our local JIRA we were having issue with Strings being set as story points this should make sure that and exception is not thrown.


collectors/feature/jira/src/main/java/com/capitalone/dashboard/client/story/StoryDataClientImpl.java
```diff
@@ -315,7 +315,7 @@ private void processFeatureData(Feature feature, Issue issue, Map<String, IssueF
 		
 		// sStoryPoints
 		IssueField storyPointsField = fields.get(featureSettings.getJiraStoryPointsFieldName());
-		if (storyPointsField != null && storyPointsField.getValue() != null && !TOOLS.sanitizeResponse(storyPointsField.getValue()).isEmpty()) {
+		if (storyPointsField != null && storyPointsField.getValue() != null && !TOOLS.sanitizeResponse(storyPointsField.getValue()).isEmpty() && isNumeric(TOOLS.sanitizeResponse(storyPointsField.getValue()))) {
 			Double value = Double.parseDouble(TOOLS.sanitizeResponse(storyPointsField.getValue()));
```
satishc1: @irvingwa Instead of checking every string for isNumeric you can catch the numberformatexception.

collectors/feature/jira/src/main/java/com/capitalone/dashboard/client/story/StoryDataClientImpl.java
```diff
@@ -754,4 +754,8 @@ private void updateStatuses() {
 			}
 		}
 	}
+	
+	private boolean isNumeric(String s) {
+    		return s != null && s.matches("[-+]?\\d*\\.?\\d+");
+  	}
 }
```
satishc1: @irvingwa not needed. Catch numberformatexception instead.

satishc1:
Fixed in https://github.com/capitalone/Hygieia/pull/1888

PR #1773, opened 2017-10-12
Hygieia - Pagination - All dashboards 

nireeshT:
This PR includes server side pagination in Hygieia landing page for All Dashboards section.
Pagination is implemented to improve performance on landing page, get results basing on page size value (default value is set to 10). Consequent clicks on next pages will pull data sets depending on page number value. You can custom set page size value can be set in api properties file by adding below property 
*** pageSize=Integer value **** 
Filter dashboards functionality is implemented - search criteria basing on dashboard title. Results will auto paginate as per page size.

screen shot 2017-10-12 at 4 15 30 pm


satishc1:
@tabladrum LGTM

PR #1775, opened 2017-10-13
Added page size property in api properties

nireeshT:
Added page size property in api properties - default value is set to 10


satishc1:
@tabladrum @Megha849 LGTM

PR #1802, opened 2017-10-29
re-upgrade docker-compose to v2

rbtr:
as noted by @kwhickey in https://github.com/capitalone/Hygieia/pull/873 the upgrade to v2 was merged, then undone by a later merge. this is an attempt to reintegrate that with some of the subsequent changes.

@charleswisniewski
@jdewinne
@virtualpeter

can you verify your intent? i have tried to capture all the changes from [here](https://github.com/capitalone/Hygieia/commits/master/docker-compose.yml) that were made after the compose got clobbered back to v1.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,349 +1,383 @@
-mongodb:
-  image: mongo:latest
-  environment:
-    - MONGODB_USERNAME=db
-    - MONGODB_DATABASE=dashboard
-    - MONGODB_PASSWORD=dashboard
+version: '2'
+volumes:
+  data: {}
+services:
+  mongodb:
+    image: mongo:latest
+    container_name: mongodb
+    #command: mongod --smallfiles
+    ports:
+     - "27017:27017"
+    volumes:
+     - data:/data/db:rw
+    environment:
+     - MONGODB_USERNAME=db
+     - MONGODB_DATABASE=dashboard
+     - MONGODB_PASSWORD=dashboard
+    volume_driver: local
+  db-setup:
+    build: db
+    depends_on: [mongodb]
+    container_name: db_doer
+    links:
+      - mongodb
+    #links now use hostnames instead of ENV Vars
+    environment:
+      - MONGO_STARTED=yes
+  hygieia-api:
+    build: api/docker
+    image: library/hygieia-api:latest
+    container_name: hygieia-api
+    ports:
+      - "8080:8080"
+    volumes:
+    - ./logs:/hygieia/logs
+    environment:
+      - MONGO_STARTED=yes
+      - jasypt.encryptor.password=hygieiasecret
+      - SPRING_DATA_MONGODB_DATABASE=dashboard
+      - SPRING_DATA_MONGODB_HOST=mongodb
+      - SPRING_DATA_MONGODB_PORT=27017
+      - SPRING_DATA_MONGODB_USERNAME=db
+      - SPRING_DATA_MONGODB_PASSWORD=dashboard
+    links:
+      - mongodb
+  hygieia-ui:
+    build: UI/docker
+    image: library/hygieia-ui:latest
+    container_name: hygieia-ui
+    ports:
+      - "8088:80"
+    links:
+      - hygieia-api
+  hygieia-github-scm-collector:
+    image: capitalone/hygieia-github-scm-collector:latest
+    container_name: hygieia-github
+    volumes:
+      - ./logs:/hygieia/logs
+    links:
+      - mongodb:mongo
+      - hygieia-api
+    #place any values you want to change in docker-compose.override.yml
+    environment:
+    - GITHUB_HOST=github.com
+    #- GITHUB_CRON='0 0/5 * * * *'
+    - GITHUB_COMMIT_THREASHOLD_DAYS=15
+    - MONGO_PORT=tcp://mongodb:27017
+  # hygieia-jenkins-build-collector:
+  #   image: capitalone/hygieia-jenkins-build-collector:latest
+  #   container_name: hygieia-jenkins-build
+  #   volumes:
+  #   - ./logs:/hygieia/logs
+  #   links:
+  #   - mongodb:mongo
+  #   - hygieia-api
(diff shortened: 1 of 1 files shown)


virtualpeter:
i think the only change relevant to me was the appdynamics collector. i have checked that and it seems ok.

overall I think its time to refactor the dockerfiles and docker-compose to support running in docker swarm or kubernetes environments and that it is time to review how env variables and the collector properties scripts are used - the pattern is working ok, but is at odds with newer docker secrets / docker config.

For now though it is a good start to move away from v1 and its implied ENV variable sharing between containers so Im in favour of this change

rbtr:
@virtualpeter thanks.
i'm happy to further upgrade this to v3.2/swarm as the next step after this is merged 😁

docker-compose.yml
```diff
@@ -1,349 +1,383 @@
...
+    volumes:
+    - ./logs:/hygieia/logs
+    environment:
+      - MONGO_STARTED=yes
+      - jasypt.encryptor.password=hygieiasecret
+      - SPRING_DATA_MONGODB_DATABASE=dashboard
+      - SPRING_DATA_MONGODB_HOST=mongodb
+      - SPRING_DATA_MONGODB_PORT=27017
+      - SPRING_DATA_MONGODB_USERNAME=db
+      - SPRING_DATA_MONGODB_PASSWORD=dashboard
```
matthewbeasley: This should be `- SPRING_DATA_MONGODB_PASSWORD=dbpass` in order to match the values found in [db-setup.js](https://github.com/capitalone/Hygieia/blob/master/db/db-setup.js).

docker-compose.yml
```diff
@@ -1,349 +1,383 @@
...
+    links:
+      - mongodb
+  hygieia-ui:
+    build: UI/docker
+    image: library/hygieia-ui:latest
```
matthewbeasley: This should probably be `image: hygieia-ui:latest` since currently we don't update the images in Docker Hub often enough

docker-compose.yml
```diff
@@ -1,349 +1,383 @@
...
+    links:
+      - mongodb:mongo
+      - hygieia-api
+    #place any values you want to change in docker-compose.override.yml
+    environment:
+    - GITHUB_HOST=github.com
+    #- GITHUB_CRON='0 0/5 * * * *'
+    - GITHUB_COMMIT_THREASHOLD_DAYS=15
+    - MONGO_PORT=tcp://mongodb:27017
+  # hygieia-jenkins-build-collector:
```
matthewbeasley: This should be removed as it is duplicated later on

docker-compose.yml
matthewbeasley: On second thought both should be changed to reflect the documentation

docker-compose.yml
```diff
@@ -1,349 +1,383 @@
...
-  - mongodb:mongo
-  - hygieia-api
-# place any values you want to change in docker-compose.override.yml
-#  environment:
-#  - APPDYNAMICS_INSTANCE_URL=http://appdynamics
-#  - APPDYNAMICS_USERNAME=username
-#  - APPDYNAMICS_PASSWORD=password
-#  - APPDYNAMICS_DASHBOARD_URL=http://appdynamics/controller/#/location=APP_DASHBOARD&timeRange=last_15_minutes.BEFORE_NOW.-1.-1.15&application=%s&dashboardMode=force
-#  - APPDYNAMICS_CRON='-1 * * * * *'
+version: '2'
```
jdewinne: Shouldn't we go for version '3'

docker-compose.yml
```diff
@@ -1,349 +1,383 @@
...
+volumes:
+  data: {}
+services:
+  mongodb:
+    image: mongo:latest
```
jdewinne: If I remember correctly this was my core change. Using the standard mongo image didn't work for me, so I had to change that.

jdewinne: Can anyone confirm using mongo:latest now properly works for them? If so, I'll give it another try.

docker-compose.yml
rbtr: if this was actually supposed to be an upgrade, i would have targeted 3. this was just trying to re-implement several other merged changes that then got clobbered by poor merge management

docker-compose.yml
matthewbeasley: Can't confirm. As of right now I haven't been able to get the db-setup to work, but haven't spent much time with it. I'm just using a separate mongo instance with a persistent data volume. Which mongo image did you use to get it to work?

matthewbeasley:
@rbtr @satishc1 - I think we should not merge this PR for two reasons:
1: we need to overhaul the docker-compose file. Currently it tries to do way too much. I think it should be split up into a "core" file which creates and connects the UI, api, and db and then each collector should have it's own file that extends the core file. 
2: this PR is sloppy. It has white space issues, inconsistent image naming, and development trash is left in
I get that this wasn't supposed to be an upgrade and was just intended to recapture the 'clobbered' changes but I can't imagine merging this one without a follow up PR and I don't think that's worth the time when we really need to change the entire file in the near future anyway.

satishc1:
@matthewbeasley Agreed. No point merging a sloppy pr done just for the sake of recapturing  'clobbered' changes. Thanks.

rbtr:
@matthewbeasley 
to 1.  If you want to say that this isn't worth merging because the whole compose file needs to be scrapped and redone, fine, but let's be clear that this changeset only puts things back have been removed accidently. 

Which brings me to 2.  I evidently didn't make myself clear enough. I didn't write any code in this PR. All of this was already in master at several points in the history. This cleanup was done just in git, to consolidate and put back changes that had already been **approved and merged** but were then undone by a later merge with an earlier root commit.

matthewbeasley:
@rbtr - believe me, you’ve made it very clear that you didn’t do any original work in this PR. We really appreciate your interest in this project and if you would like to submit a separate PR that targets compose v3 or even recaptures the changes cleanly we would love to review that, but as stated above we don’t feel that it is worth it to merge this one.

PR #1805, opened 2017-10-30
changes for adding POST endpoint to add Perf results and Perf Audit API endpoint

sashi-kan:
Adds a new POST endpoint "/quality/perftest" which consumes a "PerfTestDataCreateRequest.java" model to post Perf results in Cucumber Json format.

The new endpoint allows 3rd party performance testing applications to post KPI (Key Performance indicator) metrics which out the need for a Jenkins build.

"runId" property on PerfTestDataCreateRequest should be unique for every test execution. 
Unique names for "testName" and "perfTool" properties will create a new collector Item, under which a new Test result record would be added.

11/13/2017-- Audit API work added along with audit status


api-audit/src/main/java/com/capitalone/dashboard/service/AuditServiceImpl.java
```diff
@@ -946,11 +904,93 @@ public boolean isGitRepoConfigured(String url, String branch) {
...
+
+    private List<TestResult> getTestResults(String jobUrl, long beginDt, long endDt) {
+        List<TestResult> testResults = customRepositoryQuery
+                .findByUrlAndTimestampGreaterThanEqualAndTimestampLessThanEqual(jobUrl, beginDt, endDt);
+
+        return testResults;
+    }
+
+    public PerfReviewResponse getresultsBycomponetAndTime(String businessComp, long from, long to) {
+        Cmdb cmdb = cmdbRepository.findByConfigurationItemIgnoreCase(businessComp); // get CMDB iD
```
nireeshT: @sashi-kan  When I run GET validatePerfResults with businessComp name which is invalid, I get below error with nullpointer exception. Please add null check and add a message "Invalid businessComp name or does not exists" in return response. 

{"timestamp":1513021211490,"status":500,"error":"Internal Server Error","exception":"java.lang.NullPointerException","message":"No message available","path":"/apiaudit/validatePerfResults"}

api-audit/src/main/java/com/capitalone/dashboard/service/AuditServiceImpl.java
```diff
@@ -2,26 +2,7 @@
 
 import com.capitalone.dashboard.ApiSettings;
 import com.capitalone.dashboard.misc.HygieiaException;
-import com.capitalone.dashboard.model.AuditStatus;
```
nireeshT: @sashi-kan Please do not reformat entire file. Add your changes to existing format , that way it will just show up your changes. Add your imports to existing instead of import pakage

satishc1:
@sashi-kan @nireeshT Thank you both.

satishc1:
@tabladrum LGTM

PR #1807, opened 2017-10-31
Introduce CHANGELOG to keep track of changes

amaltson:
Following the [Keep a Changelog format](http://keepachangelog.com/en/1.0.0/), introduce a CHANGELOG to keep track of changes to the project. Unfortunately the tagging in GitHub only captures version `2.0.2` and `2.0.0` (and older 1.x versions), so it's a little hard to see what's changed in-between.

Therefore, for now I've added the sections themselves to show how the CHANGELOG works and I hope we can work together to add tagging and flush out high level details to the CHANGELOG that we can communicate to end users on what has changed.

Hope this is useful! Thanks!


satishc1:
@tabladrum LGTM

PR #1812, opened 2017-11-02
Fix commit type for squash merged and rebase merged PRs

satishc1:
Fix commit type for squash merged and rebase merged PRs. They currently show up as New commits instead of Merge. This will not fix old commits that exist in the database already. Those will still show up as type new unless you do a data cleanup.


satishc1:
@tabladrum LGTM

PR #1888, opened 2017-12-06
Story Points Field Value: Logic to handle invalid (non-number) values

subodhbattina:
Hi Satish,
Have added the logic to handle and log invalid (non-number) values for Story Point Field Values.
And yes, I think handling the NumberFormatException is better way of doing it, that way it lets us log an error message as well. Thanks !


satishc1:
@tabladrum @subodhbattina LGTM

Issue #1856, opened 2017-11-20
API_HOST is hardcoded in Hygieia/UI/docker/conf-builder.sh


ibnjay:
We have two Hygieia env deployed in our docker infra. This hardcoded value in conf-builder.sh prevents us from being able to provide different API_HOST env variable for the two deployment.

matthewbeasley:
Hey ibnjay! I'm currently looking into this. Are you using legacy container linking?

ibnjay:
@matthewbeasley Sorry didn't see that i already had a response. No not using legacy linking. Basically We launch the stack using docker-compose. Both instance of hygeia use the name network to expose frontend URLs. Using secrets we can configure API_HOST env for both stack differently. But the conf-builder.sh hard codes the value instead of just using the env variable. I suggest adding a check ```if API_HOST != null then export API_HOST=hygieia-api ``` Same thing for port.

PR #1907, opened 2017-12-15
UI conf-builder.sh to have sensible default for API_HOST

matthewbeasley:
This solves issue [#1856](https://github.com/capitalone/Hygieia/issues/1856). In the current state no matter what you attempt to set API_HOST to, you always get 'hygieia-api' as the hostname. In addition 127.0.0.1 is the current default which is rather useless as the api will never be reachable at this address.

The solution: remove the two export statements because they interfere with a user's ability to set the api hostname using ENV vars. Also change '127.0.0.1' to 'hygieia-api' since that hostname will be provided for you if you are using container linking or networking.

### DIFF
--- a/UI/docker/conf-builder.sh
+++ b/UI/docker/conf-builder.sh
@@ -1,10 +1,4 @@
 #!/bin/bash
 
-
-# if we are linked, use that info
-#linked containers now use hostnames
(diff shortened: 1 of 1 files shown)


satishc1:
@tabladrum LGTM

PR #1920, opened 2018-01-05
ng-if instead of ng-show to prevent duplicated ids

matthewbeasley:
The login page uses ng-show to hide / show the login forms for standard and ldap. This is problematic because both are loaded into the DOM. This is because ng-show loads the element into the DOM and then uses the condition provided to determine an elements visibility. If we use ng-if instead then the elements are only loaded into the DOM if the condition is met.


matthewbeasley:
image

nireeshT:
@satishc1  LGTM

PR #1921, opened 2018-01-05
Updated JIRA pom.xml so builds... build.

Preskton:
Had problems building the JIRA collector when pointing to `m2proxy.atlassian.net`, which doesn't seem to exist any longer (DNS lookup failed from multiple locations). Successor appears to be `maven.atlassian.net`. After changing, JIRA collector build was successful.


evolvah:
I came across the same problem. However, I used a different URL for the repository, the one that is hosted in Atlassian's Artifactory: `https://packages.atlassian.com/maven-public`

Which one is a better choice?

Preskton:
If you hit `maven.atlassian.com`, you land in the Artifactory home. I'm fine with either URL, though, - if it works, it works. :)

subodhbattina:
@satishc1 LGTM

Issue #1780, opened 2017-10-17
Github Collector: Use GraphQL API


tabladrum:
Seems Github GraphQL api's can lift up collector performance quite a bit. Proposing that we rewrite the Github collector to use these APIs instead of the REST APIs.

tabladrum:
See this PR: https://github.com/capitalone/Hygieia/pull/1923

tabladrum:
Closing this issue

PR #1923, opened 2018-01-07
Add github GraphQL collector and re-factor audit

tabladrum:
Added github GraphQL collector. This has several performance improvements over the github collector that uses rest api. Eventually the github collector that uses rest api will be deprecated. Also re-factored the audit api.


satishc1:
LGTM awesome xmas gift.

tabladrum:
https://github.com/capitalone/Hygieia/issues/1780

Issue #1479, opened 2017-06-15
Performance Collector NullPointer Exception


ganeshmuralidhar:
Hi,

When I configured AppDYnamics collector with a dashboard configured for Node health metrics, I am getting a Nnullpointer Exception when I run the AppD collector and no data is reflecting on Hygieia Performance widget.

Is there a specific requirement for Dashboard content on AppD?

nireeshT:
@ganeshmuralidhar  - share collector log.

ganeshmuralidhar:
here it is . Also attached is the snapshot of my appdynamics.properties file.


INFO  c.c.d.collector.CollectorTask - Running Collector: Appdynamics
2017-06-16 09:37:01,009 ERROR o.s.s.s.TaskUtils$LoggingErrorHandler - Unexpected error occurred in scheduled task.
java.lang.NullPointerException: null
        at com.capitalone.dashboard.collector.AppdynamicsSettin
![untitled](https://user-images.githubusercontent.com/29456270/27211570-cbba9856-5277-11e7-9e7c-7097c3b39ca3.png)
gs.getInstanceUrlList(AppdynamicsSettings.java:38) ~[appdynamics-performance-collector-2.0.5-SNAPSHOT.jar!/:2.0.5-SNAPSHOT]
        at com.capitalone.dashboard.model.AppdynamicsCollector.prototype(AppdynamicsCollector.java:31) ~[appdynamics-performance-collector-2.0.5-SNAPSHOT.jar!/:2.0.5-SNAPSHOT]
        at com.capitalone.dashboard.collector.AppdynamicsCollectorTask.getCollector(AppdynamicsCollectorTask.java:50) ~[appdynamics-performance-collector-2.0.5-SNAPSHOT.jar!/:2.0.5-SNAPSHOT]
        at com.capitalone.dashboard.collector.AppdynamicsCollectorTask.getCollector(AppdynamicsCollectorTask.java:20) ~[appdynamics-performance-collector-2.0.5-SNAPSHOT.jar!/:2.0.5-SNAPSHOT]
        at com.capitalone.dashboard.collector.CollectorTask.run(CollectorTask.java:48) ~[core-2.0.5-SNAPSHOT.jar!/:2.0.5-SNAPSHOT]
[...]

PR #1944, opened 2018-01-18
Check size of metricValues

ulthuan:
When we try to use AppDynamics docker collector I have problems because the script of 'property generator' creates a file with instanceUrl without the final 's' character. Also the java code has a problem when the json of api rest response doesn't have anything in the metricValues array.

The change of the 'property generator' is mentioned in issue #1479


collectors/performance/appdynamics/src/main/java/com/capitalone/dashboard/collector/DefaultAppdynamicsClient.java
```diff
@@ -160,7 +160,11 @@ public static String joinURL(String base, String... paths) throws MalformedURLEx
                 for (Object entry : array) {
                     JSONObject jsonEntry = (JSONObject) entry;
                     String metricPath = getString(jsonEntry, "metricPath");
-                    JSONObject mObj = (JSONObject) getJsonArray(jsonEntry, "metricValues").get(0);
+                    JSONArray metricValues =  getJsonArray(jsonEntry, "metricValues");
+                    if (metricValues.size() == 0){
+                        continue;
```
satishc1: Please use CollectionUtils.isEmpty(metricValues) instead

ulthuan:
Hi @satishc1!,
I did the change in the code :)

Thanks for your time!

collectors/performance/appdynamics/src/main/java/com/capitalone/dashboard/collector/DefaultAppdynamicsClient.java
```diff
@@ -160,7 +161,11 @@ public static String joinURL(String base, String... paths) throws MalformedURLEx
                 for (Object entry : array) {
                     JSONObject jsonEntry = (JSONObject) entry;
                     String metricPath = getString(jsonEntry, "metricPath");
-                    JSONObject mObj = (JSONObject) getJsonArray(jsonEntry, "metricValues").get(0);
+                    JSONArray metricValues =  getJsonArray(jsonEntry, "metricValues");
+                    if (metricValues.isEmpty()){
```
Sbrenthughes: Can you please replace with CollectionUtils.isEmpty() here?

ulthuan: I try to change, but the type of this object (JSONArray) is not compatible with this function

ulthuan: This is the reason because I try to use a similar function

satishc1:
@ulthuan @Sbrenthughes Thank you

satishc1:
@tabladrum LGTM

PR #1960, opened 2018-01-23
Fix for Perf audit API.

sashi-kan:


satishc1:
@sashi-kan @tabladrum LGTM

PR #1968, opened 2018-01-26
Adding a set of test utils to facilitate better integration testing f…

tabladrum:
…or collectors with Fake Mongo and a customizable TestRestTemplate to fake out REST response from tools.

Had to add it to core. Will see later if it makes sense to create a separate hygieia core test library.


satishc1:
@tabladrum LGTM

PR #1972, opened 2018-01-28
Docker compose refactor

matthewbeasley:
The goal of this PR is make it so that a user can run `mvn install docker:build; docker-compose up -d` and have a UI, api, and a dev db. Then the user can rename the `docker-compose.override.example.yml` to `docker-compose.override.yml`, set the environment vars for the collectors they want to run and have those come up too. The compose file also maps the port of the mongodb in order to allow you to run the api or the collectors on the host machine if you are developing them. The api port is also mapped for development of the UI. The hose names have been changed to be the set to the names of the services. Docker Compose allows you to reference other containers via service name. The db information has been changed to match the docs. 

Future plans/ thoughts: clean up the `properties-builder.sh` files. The env var names are far too long for most of them. Also the way it currently works disallows optional env vars i.e all env vars must be set or lines must be commented out of the `properties-builder.sh` file in order for the collectors or the api to start. I was thinking of doing something like this:
```
if [ -z $ENV_VAR ]
then
  echo "collector.property=$ENV_VAR" >> $PROP_FILE
fi
```
However, I couldn't get it to work in time. The `Dockerfile`s themselves need to be cleaned as well. The `MAINTAINER` tag is deprecated for example. I would also like to add all the potential environment overrides to the override file to work as a bit of a reference to what all can be achieved


satishc1:
@tabladrum LGTM

Issue #1851, opened 2017-11-20
Hygieia dashboard - Quality Widget : "VIEW ALL” doesn't take to the sonar page


lenalevy:
I have configure on Hygieia dashboard on “Configure Widget” Quality section.
I have choose Micro Service.
The Widget is presented , but once I'm clicked in "View all " option , it's open new page and I'm getting error "Componenet key <id> not found "
I'm understanding that the “VIEW ALL” should takes to the sonar page of respective MS
![soner error](https://user-images.githubusercontent.com/33826321/33020746-091b425c-ce08-11e7-9c45-b1a79a8f9dce.PNG)

nireeshT:
@lenalevy  send screenshot of quality widget and the url it is taking to when clicked on "View ALL". if you see any error in console, share it.

lenalevy:
I found the solution for this problem .
I have changed view.js under UI/src/components/widgets/codeanalysis
Fix : 
function processCaResponse(response) 

ctrl.reportUrl = response.reportUrl;       - orig code 
ctrl.reportUrl = caData.url;                          - my fix

Now it's working fine

matthewbeasley:
@lenalevy - thanks for taking the time to post the solution!

PR #1987, opened 2018-02-05
Updated the link to the sonar scan

matthewbeasley:
image

Before the change this is what the link looked like. After the change you can see that you reach the appropriate scan results. 

This solves [#1851](https://github.com/capitalone/Hygieia/issues/1851)


satishc1:
@tabladrum LGTM

PR #1989, opened 2018-02-05
Retrieve Sonar collector config changes only for sonarqube versions greater than 5.0

ReddytoRumble:
This catches all exceptions when the collector attempts to retrieve a configuration change and logs it so that it does not prevent the collector task from collecting data for any other sonar urls

It also only retrieves config changes for Sonarqube versions greater than 5.0

@satishc1


satishc1:
@tabladrum LGTM

PR #1993, opened 2018-02-07
Updated timestamps in models used by hpsm collector.

DvdNelson:
Updated timestamps in models used by hpsm collector (Cmdb, ChangeOrder, Incident) to use Long for storage.  Re-added and corrected String set methods for timestamps


satishc1:
@tabladrum LGTM

PR #2015, opened 2018-02-12
Remove dependency on Object Ids 

Sbrenthughes:
PR is for removing the dashboards dependency on configuration Item Object Ids. It places the Dependency on the configuration item names instead.


satishc1:
@tabladrum LGTM

PR #2017, opened 2018-02-12
Build widget fix - Search with job name

nireeshT:
Build widget fix - Search with job name

PR #2020, opened 2018-02-12
Revert "Build widget fix - Search with job name"

satishc1:
Reverts capitalone/Hygieia#2017


satishc1:
@tabladrum @nireeshT Reverting. Please investigate the possibility of moving the queries to collectorItemRepository. Also we want to ensure its not tool specific..ie. it should work for any build tool.

PR #2024, opened 2018-02-13
Code Quality Widget - Disable "View All" if no scan has been performed yet

matthewbeasley:
If no scan data is present in the db then the "View All" button on the Code Quality widget will set an empty href. In this case when you click on the button it will take you the home page since the Hygieia UI is a single page app. This PR uses ng-href instead which supports expressions. So now if there is no scan data then caWidget.reportUrl will evaluate to false and no href will be set.


nireeshT:
@satishc1  @tabladrum  LGTM

PR #2058, opened 2018-02-27
Modify error handling

satishc1:
Changes a couple of things:
1. Adds repo_not_configured and pending_data_collection statuses to codereviewaudit.
2. Doesn't clear out errors when the remoteUpdate api updates a collector item.
3. Changes the reset logic for a collector item where instead of clearing all errors, it only clears the oldest one so as to drop it below the threshold count.


satishc1:
@tabladrum LGTM

PR #2059, opened 2018-02-27
Repo widget - Get SCM type from collectorItems

nireeshT:
This PR adds changes to get scm type from collectorItems instead from widgetConfig.


satishc1:
@tabladrum LGTM

PR #2065, opened 2018-03-06
Fully working docker-compose setup

ianrossi:
@matthewbeasley @nireeshT 
This PR:
* represents a fully working, "push-button" docker-compose setup
* moves away from the deprecated [docker-maven plugin](https://github.com/spotify/docker-maven-plugin) and uses [Spotify's newer dockerfile-maven plugin](https://github.com/spotify/dockerfile-maven).
* adds a Vagrant setup for development and testing purposes.

The reason for the PR is that I could not get the docker-compose configuration to work out of the box.

# Usage
Basically, you just put your `docker-compose.override.yml` file in the repo's root dir and run the `provision_script.sh` shell script. It will install Docker, perform the Maven/Docker build and then run `docker-compose` based on the contents of `docker-compose.override.yml`.

This is all tested and works.

There are 2 use cases:
1. On a standalone server, where you just perform the steps described above.
2. A Vagrant setup, where you put your `docker-compose.override.yml` file in place in the repo root, install Vagrant, then execute `vagrant up` from the repo root. This will provision a Vagrant box in Virtualbox and then perform install, build and deploy in Docker.


mikesalvia:
Many conflicts which are divergent from the current method of `mvn docker: build` probably can be closed.

nireeshT:
@ianrossi  Thanks for contribution. we are working on it, will get back to you soon.

matthewbeasley:
@ianrossi as I mentioned on gitter.im, I have a few concerns: The first is the conflicts due to me having a unmerged PR at the time you forked Hygieia that has since been merged. The second is that I'm concerned with binding to the default maven life cycle phases. As it stands, I don't believe the containers are ready to be the de facto artifact for production. Finally I don't think `--allow-root` should be in your UI/pom.xml. Just chown those files. 

Of course, we can talk about it. We are definitely interested in these changes.

PR #2073, opened 2018-03-13
UI changes for General Configuration

dhivahar19:
Configuration page for collectors server configuration details

PR #2074, opened 2018-03-13
Configuration option in admin page with collector changes & Rally collector initial commit

Satheesh-Balachandran:
Allows the admin to configure the servers and credentials with which the collectors will fetch the data. Collector changes for Nexus-iq-collector and Jenkins-build collector.


nireeshT:
@Satheesh-Balachandran  please upload screenshots for this change

Satheesh-Balachandran:
hygieia configuration page
@nireeshT These are Java changes for the PR #2073.

core/src/main/java/com/capitalone/dashboard/model/Configuration.java
```diff
@@ -0,0 +1,61 @@
...
+
+	private final static String PREFIX = "BASE64:";
+
+	public Configuration(String collectorName, Set<Map<String, String>> info) {
+		super();
+		this.collectorName = collectorName;
+		this.info = info;
+	}
+
+	public Set<Map<String, String>> decryptOrEncrptInfo() {
```
nireeshT: @Satheesh-Balachandran  can we use existing Encryption utils for encryption and decryption. That way it is easy to secure confidential information

nireeshT:
@Satheesh-Balachandran  Can you detail how we can use this configuration once it is saved to Mongo. I see a record generated in Configuration collection, what's the purpose of that configuration item.

core/src/main/java/com/capitalone/dashboard/model/Configuration.java
Satheesh-Balachandran: @nireeshT Encryption utils required api key to encrypt data. Since, this method will be used by the collectors to decrypt the password to fetch data from tools, api key will not be available here. So its better to inject encryption here and also not to repeat code in each collectors for decryption.

Satheesh-Balachandran:
The config data(url, username, password) will be fetched from each collectors on each run. If the password for a particular tool, say rally, changes at any time, it need not have to be changed in the application.properties. The admin can just change this in the configuration page UI(#2073 ) and without restarting the collector, the changes will be adapted automatically.

nireeshT:
@Satheesh-Balachandran  I see this as a good baseline for updating properties for all collectors.Sounds good

nireeshT:
@dhivahar19  @Satheesh-Balachandran  please update branch. We are planning to merge this PR

Sbrenthughes:
@Satheesh-Balachandran Can you update branch and re push?

Issue #2028, opened 2018-02-14
CodeQualityCollector not working


josefarinna:
Hi,

I was using this collector for a few months with SonarQube 5.6 LTS, now we moved to SonarQube 6.7.1 LTS and seems that is not able to parse the response from the server. I also tried with an SonarQube 7.0 and I'm getting a similar error.

The exception:
```
2018-02-14 11:15:00,015 INFO  c.c.d.collector.CollectorTask - Running Collector: Sonar
2018-02-14 11:15:00,171 INFO  c.c.d.collector.CollectorTask - -----------------------------------
2018-02-14 11:15:00,171 INFO  c.c.d.collector.CollectorTask - http://10.247.0.25:9000
2018-02-14 11:15:00,171 INFO  c.c.d.collector.CollectorTask - -----------------------------------
2018-02-14 11:15:00,781 INFO  c.c.d.collector.CollectorTask - Fetched projects   6             0s
2018-02-14 11:15:00,812 INFO  c.c.d.collector.CollectorTask - New projects            6        0s
2018-02-14 11:15:00,874 INFO  c.c.d.collector.CollectorTask - Updated                 0        0s
2018-02-14 11:15:03,126 ERROR c.c.d.collector.DefaultSonar6Client - Could not parse response from: http://10.247.0.25:9000/api/qualityprofiles/changelog?profileKey=js-sonar-way-recommended-50412
org.json.simple.parser.ParseException: null
	at org.json.simple.parser.Yylex.yylex(Yylex.java:610) ~[json-simple-1.1.1.jar!/:na]
	at org.json.simple.parser.JSONParser.nextToken(JSONParser.java:269) ~[json-simple-1.1.1.jar!/:na]
	at org.json.simple.parser.JSONParser.parse(JSONParser.java:118) ~[json-simple-1.1.1.jar!/:na]
	at org.json.simple.parser.JSONParser.parse(JSONParser.java:81) ~[json-simple-1.1.1.jar!/:na]
	at org.json.simple.parser.JSONParser.parse(JSONParser.java:75) ~[json-simple-1.1.1.jar!/:na]
	at com.capitalone.dashboard.collector.DefaultSonar6Client.parseAsArray(DefaultSonar6Client.java:218) ~[sonar-codequality-collector-2.0.5-SNAPSHOT.jar!/:2.0.5-SNAPSHOT]
[...]

PR #2092, opened 2018-03-26
Fix sonar CodeQuallity collector parse error (#2028)

josefarinna:
This commit fix the issue #2028


nireeshT:
@josefarinna  is this fix backwards compatible for Sonar 5.3 ?

josefarinna:
Hello @nireeshT, this fix is only applicable to versions 6.3 and up (DefaultSonar6Client.java is for 6.3 and up), for previous versions this doesnt affect. But to be sure, I deployed and old version of sonarqube in my machine and used it to fetch the projects, and everything was working fine.

Issue #1599, opened 2017-08-08
BUG: Cannot access test data in UI


htvan1:
Data does seem to get loaded from what I can tell when watching the calls back and forth. But something gets weird if you try and log in or create a user when using:  gulp serve --local true
Also the documentation does not state what the password is for user created in the test data: someuser

When it tries to authenticate on a log in, the header is missing. response.headers()['x-authentication-token'])  ...in fact no headers are set, the response looks completely different from the one you will see when you do not load with --local true. 
Logging in and creating a new user works fine with one's own data.

TypeError: response.headers is not a function
    at processResponse (authService.js:15)
    at processQueue (bower.js:28357)
    at bower.js:28373
    at Scope.$eval (bower.js:29655)
    at Scope.$digest (bower.js:29469)
    at Scope.$apply (bower.js:29763)
    at done (bower.js:23743)
    at completeRequest (bower.js:23952)
    at XMLHttpRequest.requestLoaded (bower.js:23880)
(anonymous) @ bower.js:25860
(anonymous) @ bower.js:22368
processQueue @ bower.js:28365
(anonymous) @ bower.js:28373
$eval @ bower.js:29655
$digest @ bower.js:29469
$apply @ bower.js:29763
done @ bower.js:23743
completeRequest @ bower.js:23952
requestLoaded @ bower.js:23880
XMLHttpRequest.send (async)
(anonymous) @ bower.js:23926
sendReq @ bower.js:23688
serverRequest @ bower.js:23483
processQueue @ bower.js:28357
(anonymous) @ bower.js:28373
$eval @ bower.js:29655
$digest @ bower.js:29469
$apply @ bower.js:29763
(anonymous) @ bower.js:38258
dispatch @ bower.js:5206
elemData.handle @ bower.js:5014

Sbrenthughes:
https://github.com/capitalone/Hygieia/issues/1554 try looking at this.

htvan1:
What nireeshT recommend  doesn't fix general issue with the test data provided, that does fix auth log in but not for --local true.  If I removed the true boolean from the log in/auth pages, this header issues goes away.  I log in with a user I have made and then new errors pop up.  I believe the data model for test data is not current or has not been updated to reflect changes at some point.  I worked through another data model issue, but then another one just popped up...So I just stopped trying to use it.  I believe the test data needs a close inspection and refreshing to match the current code base.

htvan1:
[...]

PR #2094, opened 2018-03-26
[Fix] TypeError: Cannot read property 'name' of undefined

jozzhart:
https://github.com/capitalone/Hygieia/issues/1599#issuecomment-326066852


Sbrenthughes:
@jozzhart Can you update and push again?

PR #2107, opened 2018-03-29
Set DB properties to the correct defaults in all collectors

spennec:
Some collectors have the correct values but most don't. I took the values from the DB setup script and modified all the properties-builder.sh to use them, so we don't have to provide those in the docker-compose.override.yml for each of the collectors.


spennec:
oops sure, will do. Sorry for the mess.

spennec:
It's all good now, there's no modifications on the score docker file in this PR.

Sbrenthughes:
@spennec Can you update your branch and re push.

spennec:
@Sbrenthughes updated it directly via GitHub, seems ok.

Issue #2089, opened 2018-03-23
Permission denied when connecting to UI


spennec:
Hello,

I'm making a POC with Hygieia, and I've had the following problem: when I call the UI in a browser (http://myhost:3000), I get a 403 Forbidden error. In the UI docker image, I see many of the following logs:

2018/03/23 11:08:34 [error] 9#9: *1 open() "/usr/share/nginx/html/favicon.ico" failed (13: Permission denied), client: x.y.z., server: localhost, request: "GET /favicon.ico HTTP/1.1", host: "remotehost:3000", referrer: "http://remotehost:3000/"

I build everything straight from Github's master branch, on a RHEL box. I build the docker images, by following this documentation: http://capitalone.github.io/Hygieia/builddocker.html

After a bit of surfing, it seems that the problem is that nginx's user doesn't have the correct rights on the served files: https://github.com/docker-library/docs/issues/883

I tried several chown and chmod but it seems not to work... 

Can anybody help?

Thanks a lot,

Sébastien

matthewbeasley:
@spennec - this is typically a cookie issue. Can you go to chrome://settings/siteData - search for localhost and delete that cookie.

matthewbeasley:
@spennec - when you search make sure you are searching the cookies and not the settings. That part trips me up fairly often

spennec:
That solved the issue, thanks a lot! Now bumping into another problem but probably totally unrelated, so I'll open another issue.

spennec:
Ouch sorry: the problem was actually solved by two things:
- cleaning site data
- modifying the UI's dockerfiles (UI/Dockerfile and UI/docker/Dockerfile) to include this line:

chown -R nginx:nginx /usr/share/nginx/html/

@matthewbeasley  do you think this should be corrected in master?

matthewbeasley:
@spennec - I'm not sure that's necessary. Shouldn't the nginx image already be the owner of that directory?

spennec:
What I've seen is that the directory is owned by nginx but all the files inside belong to root. Changing them to nginx seems to solve the problem. Didn't want to open a PR for this because it's my first time using nginx, so I might be wrong.

matthewbeasley:
@spennec - yea that totally makes sense. Mind submitting another PR to add the chown to both files?

spennec:
[...]

PR #2108, opened 2018-03-29
Chown in UI dockerfiles to change owner on nginx files

spennec:
Related to issue #2089

I get a permission denied error from nginx when running the image. Clearing the site data or cookies it not sufficient. If I open a bash in the running image, I can see that the html folder is owned by nginx but all its files belong to root. Chowning them to nginx solves the problem.


UI/Dockerfile
```diff
@@ -4,6 +4,7 @@ COPY docker/default.conf /etc/nginx/conf.d/default.conf.templ
 COPY docker/conf-builder.sh /usr/bin/conf-builder.sh
 COPY dist /usr/share/nginx/html
 RUN chown nginx:nginx /usr/share/nginx/html
```
matthewbeasley: This line is now unnecessary. Please remove it.

PR #2111, opened 2018-04-02
add --allow-root in bower cmd.

leopardxu:


matthewbeasley:
@leopardxu you don't need root. Makes yourself owner of those files.

PR #2119, opened 2018-04-05
Dockerfile has previous reference to score-properties-builder.sh

bainss:


bainss:
filename change not reflected in the Jenkinsfile

PR #2122, opened 2018-04-06
Variables and their default are separated with :

spennec:
Minor typo, but makes the collector crash at startup when the property is not set in the docker-compose override.


nireeshT:
@Sbrenthughes  LGTM

PR #2123, opened 2018-04-06
Correct parsing of the MONGO_PORT variable

spennec:
The script tries to parse the wrong variable, thus never setting the correct host to the db.


spennec:
Sure! 😊 

In this collector, the properties script expects the full url of the mongo db to be passed in the MONGO_PORT variable. The comment shows the expected format:

# Sample: MONGO_PORT=tcp://172.17.0.20:27017

... but it then parses both MONGO_PORT and MONGO_HOST variables. If one follows the comment and passes the full url in the MONGO_PORT variable, then the MONGO_HOST will be empty and the script will end up with an empty mongo host to try to connect to. Or you will need to pass the full url in both variables so that each parsing extracts the correct data.

This is also done in other collectors. See the Gitlab collector's properties [script](https://github.com/capitalone/Hygieia/blob/master/collectors/scm/gitlab/docker/properties-builder.sh
) for instance.

This PR makes this script behave like the other ones, for the sake of coherence in the collectors' configs.

If I may add my 2 cents, I think that it's not a very good way to propagate the url to the docker image. Ideally, the collectors would be configured differently, like the the Sonar collectors whose [script](https://github.com/capitalone/Hygieia/blob/master/collectors/build/sonar/docker/properties-builder.sh) expects two variables for the mongo url: one with the host and another one with the port. That's probably how it should be done.

matthewbeasley:
@spennec - sonar collector is our target state. I'm going to test this PR today unless you want to go ahead and update this collector to match sonar.

spennec:
I’ve started to work on supporting custom certs in the bitbucket collector, so if it’s ok with you please test and merge this...

PR #2162, opened 2018-04-19
Centralized entreprise certificate management + addition in Bitbucket collector

spennec:
This PR provides the following feature:

- one centralized directory where one can put their entreprise certs
- declaration of a volume in docker-compose.yml that can be used in docker-compose.override
- readme file
- git ignore to make sure nobody uploads their certs to the public repo

One can attach the certs volume to any collector to make the crt available to the collector. Then, reuse the script block in Bitbucket's properties-builder.sh to add all the certs to the image's keystore.


spennec:
Hi @matthewbeasley, can you take a look at this when you have the time?

I've used this successfully on my poc instance, and believe it would help enterprise adoption, because there's a good chance that companies have internal certs to secure access to their tools.

spennec:
Hi @matthewbeasley any news about this?

PR #2165, opened 2018-04-19
Fix for Target and Achieved

sashi-kan:
Testcase steps array have data in the following order Target, achieved. Currently the code is always updating Achieved property with "0" and target property with the achieved value. This fix should set target property with target value and achieved property with achieved value.


sashi-kan:
@tabladrum @nireeshT : Could you please review?

Issue #2154, opened 2018-04-18
Missing properties in properties-builder scripts


spennec:
Hello,

I just linked a Hygieia instance to our internal bitbucket server. To do this, I add to modify some properties-builder scripts. Typically, when using private repos, both API and the collector need to share the encryption key:

- in api/docker/properties-builder.sh:
key=${KEY:-}

- in collectors/scm/bitbucket/docker/properties-builder.sh
git.key=${BITBUCKET_KEY:-}

I think that those should be permanently available in those scripts. What do you think? I can create a PR for this if you agree.

Cheers,

Sébastien

matthewbeasley:
@spennec - yes I think those should be available permanently.

PR #2168, opened 2018-04-24
Missing properties in properties-builder scripts See #2154

spennec:
Some properties are missing from two properties-builder scripts. Discussion at https://github.com/capitalone/Hygieia/issues/2154


spennec:
wow that was quick 😊 Thanks!

PR #2264, opened 2018-06-05
add in optional filters into describe instances for aws cloud collector + region support

stevegal:
it may be that you do not want to look at all instances in an account,
but a smaller more focussed list. To do this you can now use filtering on
the aws side to describe only the instances that match your filter.

### DIFF
--- a/UI/.gitignore
+++ b/UI/.gitignore
@@ -3,6 +3,9 @@
 .tern-project
 .classpath
 
+# npm files
+package-lock.json
+
 # Logs
 logs
 *.log
--- a/UI/pom.xml
+++ b/UI/pom.xml
@@ -89,7 +89,7 @@
             </goals>
             <!-- Optional configuration which provides for running any npm command -->
             <configuration>
-              <arguments>install</arguments>
+              <arguments>install --ignore-scripts</arguments>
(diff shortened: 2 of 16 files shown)


stevegal:
Also fixes test location so it is picked up by maven

nireeshT:
@stevegal  Please submit documentation for aws properties file change and possible values. 

https://github.com/capitalone/Hygieia/blob/gh-pages/pages/hygieia/collectors/cloud/aws.md

stevegal:
Will do. I’ll sort that out tomorrow. Thanks.

stevegal:
added see https://github.com/capitalone/Hygieia/pull/2269

nireeshT:
@stevegal  is this change backward compatible

stevegal:
@nireeshT should be from a code perspective as the filters are optional, and, if not provided, you’ll end up with the current functionality. I’ll double check though and get back to you.

stevegal:
@nireeshT confirmed that its all backwardly compatible.

stevegal:
@nireeshT I've also taken the opportunity to add support for non-default regions. This again is optional and will work without setting the property just as before.

Issue #2265, opened 2018-06-06
Problems build api-audit with some jvms 


stevegal:
Using Java 1.8 on a Mac, I cannot build api-audit. The model package contains a
‘’’
    public Map<AuditType, Evaluator> evaluatorMap() {
        return Collections.unmodifiableMap(Stream.of(
                new SimpleEntry<>(AuditType.CODE_REVIEW, codeReviewEvaluator),
                new SimpleEntry<>(AuditType.CODE_QUALITY, codeQualityEvaluator),
                new SimpleEntry<>(AuditType.STATIC_SECURITY_ANALYSIS, staticSecurityAnalysisEvaluator),
                new SimpleEntry<>(AuditType.LIBRARY_POLICY, libraryPolicyEvaluator),
                new SimpleEntry<>(AuditType.BUILD_REVIEW, buildEvaluator),
                new SimpleEntry<>(AuditType.TEST_RESULT, regressionTestResultEvaluator),
                new SimpleEntry<>(AuditType.PERF_TEST, performanceTestResultEvaluator))
                .collect(Collectors.toMap(SimpleEntry::getKey, SimpleEntry::getValue)));
    }’’’
That is causing a generics issue. If you use a more simple map, it build fine.

matthewbeasley:
@stevegal - which map have you tried?

stevegal:
I've fone a PR (https://github.com/capitalone/Hygieia/pull/2266) with the type 1 fix

Sbrenthughes:
Closing this as you resolved the issue and created a pr

PR #2267, opened 2018-06-06
an alternative solution to the fixed map problem without using streams

stevegal:


rvema:
@stevegal not sure if we need this PR, #2265 should be good enough, Please let know otherwise.

stevegal:
That's fine. Sometimes simple is best, Map creation doesn't need to be streamed just because it's there. One bug to track is fine

PR #2282, opened 2018-06-12
Added logic to generate collector items for hpsm incidents

subodhbattina:
Added logic to generate collector items for hpsm incidents

### DIFF
--- a/collectors/cmdb/hpsm/src/main/java/com/capitalone/dashboard/collector/DefaultHpsmClient.java
+++ b/collectors/cmdb/hpsm/src/main/java/com/capitalone/dashboard/collector/DefaultHpsmClient.java
@@ -782,7 +782,7 @@ public class DefaultHpsmClient implements HpsmClient {
 		incident.setOpenTime(getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_OPEN_TIME));
 		String closedTime = getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_CLOSE_TIME);
 		if (!StringUtils.isEmpty(closedTime)) {
-			incident.setClosedTime(getStringValueFromMap(map, HpsmCollectorConstants.INCIDENT_CLOSE_TIME));
+			incident.setClosedTime(closedTime);
 		}
 		incident.setOpenedBy(getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_OPEN_BY));
 		incident.setUpdatedTime(getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_UPDATE_TIME));
--- a/core/src/main/java/com/capitalone/dashboard/repository/DashboardRepository.java
+++ b/core/src/main/java/com/capitalone/dashboard/repository/DashboardRepository.java
@@ -43,8 +43,8 @@ public interface DashboardRepository extends PagingAndSortingRepository<Dashboar
 	Iterable<Dashboard> findAllByConfigurationItemBusServNameAndConfigurationItemBusAppName(String appName, String compName);
 
 	Dashboard findByConfigurationItemBusServNameIgnoreCaseAndConfigurationItemBusAppNameIgnoreCase(String appName, String compName);
-	List<Dashboard> findAllByTypeContainingIgnoreCaseAndConfigurationItemBusServNameContainingIgnoreCase(DashboardType type, String appName);
-	List<Dashboard> findAllByTypeContainingIgnoreCaseAndConfigurationItemBusServNameContainingIgnoreCaseAndConfigurationItemBusAppNameContainingIgnoreCase(DashboardType type, String appName, String compName);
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/capitalone/dashboard/util/DashboardUtils.java
```diff
@@ -2,15 +2,38 @@
 
 import com.capitalone.dashboard.model.Collector;
```
tabladrum: Why do we need the utils class?

subodhbattina: I had to move one method out into utils so it can be leveraged by the logic that links the collector items to the dashboard. This is to avoid code duplication.

collectors/cmdb/hpsm/src/main/java/com/capitalone/dashboard/collector/DefaultHpsmClient.java
```diff
@@ -779,6 +780,10 @@ private void handleChangeSoapMessage(SOAPBodyElement keysTag) throws SOAPExcepti
 		incident.setIncidentID(getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_ID));
 		incident.setCategory(getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_CATEGORY));
 		incident.setOpenTime(getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_OPEN_TIME));
+		String closedTime = getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_CLOSE_TIME);
+		if (!StringUtils.isEmpty(closedTime)) {
```
Sbrenthughes: Why do we need StringUtils here?

core/src/main/java/com/capitalone/dashboard/repository/DashboardRepository.java
```diff
@@ -43,6 +43,8 @@
 	Iterable<Dashboard> findAllByConfigurationItemBusServNameAndConfigurationItemBusAppName(String appName, String compName);
 
 	Dashboard findByConfigurationItemBusServNameIgnoreCaseAndConfigurationItemBusAppNameIgnoreCase(String appName, String compName);
+	List<Dashboard> findAllByTypeContainingIgnoreCaseAndConfigurationItemBusServNameContainingIgnoreCase(DashboardType type, String appName);
+	List<Dashboard> findAllByTypeContainingIgnoreCaseAndConfigurationItemBusServNameContainingIgnoreCaseAndConfigurationItemBusAppNameContainingIgnoreCase(DashboardType type, String appName, String compName);
```
Sbrenthughes: Why do you need to specify type here? Only Team dashboards can have a BusApp/ BusSer. Can `findAllByConfigurationItemBusServNameAndConfigurationItemBusAppName` not be updated to add ignoreCase and used instead?

core/src/main/java/com/capitalone/dashboard/util/DashboardUtils.java
Sbrenthughes: @tabladrum are you good here?

collectors/cmdb/hpsm/src/main/java/com/capitalone/dashboard/collector/DefaultHpsmClient.java
subodhbattina: I am just using StringUtils isEmpty method to check if the string is empty. In certain cases "closedTime" turns out to be empty.

core/src/main/java/com/capitalone/dashboard/repository/DashboardRepository.java
subodhbattina: When we create a production dashboard, there is an option on the UI to assign an Business Service, and therefore used type to make sure we are reading only Team dashboards

tabladrum: You dont need "ignoreCase" wit type here.

collectors/cmdb/hpsm/src/main/java/com/capitalone/dashboard/collector/DefaultHpsmClient.java
rvema: @subodhbattina Since you already retrieved closedTime in line 783 why cant that be set in line 785.

subodhbattina: @rvema Good catch thanks, this was by oversight, have fixed it.

core/src/main/java/com/capitalone/dashboard/repository/DashboardRepository.java
subodhbattina: @tabladrum You are right, I will remove that. Thanks

PR #2298, opened 2018-06-18
update api docker defaults

jpeerz:


matthewbeasley:
@jpeerz - The defaults provided are to be used with the docker-compose set up. The `db` hostname is provided by compose and `dbpassword` is the password from the documentation. This keeps things consistent for folks trying to set this up. If you want to provide a different password then set the `SPRING_DATA_MONGODB_HOST` environment variable.

jpeerz:
@matthewbeasley agreed, I was doing some webhook test, I'll close. cheers.

PR #2316, opened 2018-06-21
Changes to generate collector items for the incidents

subodhbattina:
Changes to generate collector items for the incidents

### DIFF
--- a/collectors/cmdb/hpsm/src/main/java/com/capitalone/dashboard/collector/DefaultHpsmClient.java
+++ b/collectors/cmdb/hpsm/src/main/java/com/capitalone/dashboard/collector/DefaultHpsmClient.java
@@ -783,9 +783,7 @@ public class DefaultHpsmClient implements HpsmClient {
 		String closedTime = getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_CLOSE_TIME);
 		if (!StringUtils.isEmpty(closedTime)) {
 			incident.setClosedTime(closedTime);
-		} else {
-            incident.setClosedTime("0");
-        }
+		}
 		incident.setOpenedBy(getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_OPEN_BY));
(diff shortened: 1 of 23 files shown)


collectors/cmdb/hpsm/src/main/java/com/capitalone/dashboard/collector/DefaultHpsmClient.java
```diff
@@ -783,6 +782,8 @@ private void handleChangeSoapMessage(SOAPBodyElement keysTag) throws SOAPExcepti
 		String closedTime = getStringValueFromMap(map,HpsmCollectorConstants.INCIDENT_CLOSE_TIME);
 		if (!StringUtils.isEmpty(closedTime)) {
 			incident.setClosedTime(closedTime);
+		} else {
```
rvema: (782 - 788) If else can be written as incident.setClosedTime((!StringUtils.isEmpty(closedTime)) ? closedTime : 0L);

core/src/main/java/com/capitalone/dashboard/model/Incident.java
```diff
@@ -79,9 +79,9 @@
 
     public void setStatus(String status) { this.status = status; }
 
-    public long getClosedTime() { return closedTime; }
+    public Long getClosedTime() { return closedTime; }
 
-    public void setClosedTime(long closedTime) { this.closedTime = closedTime; }
+    public void setClosedTime(Long closedTime) { this.closedTime = closedTime; }
 
     public void setClosedTime(String closedTime) { this.closedTime = DATE_FORMATTER.parseMillis(closedTime); }
```
rvema: public void setClosedTime(String closedTime) { this.closedTime = DATE_FORMATTER.parseMillis(closedTime); 
} 
should be written as
public void setClosedTime(String closedTime) { this.setClosedTime( DATE_FORMATTER.parseMillis(closedTime)); 
}

subodhbattina: @rvema : Sure. no problem. Yes, this is an existing overloaded method, and so we can make it use the other one. : ). Will change it out. Thanks.

Issue #1439, opened 2017-05-31
LDAP Setup


kavitha0624:
Hi Team,

I took the latest code from the Master branch and trying to setup LDAP. I have added the properties as given below. I do not see any errors in the api console but when I try to login thru LDAP, I get an error msg saying Invalid username and password.

API Properties File:

dbname=dashboardfive
dbusername=dashboarduser
dbpassword=dbpassword
dbhost=localhost
dbport=27017
dbreplicaset=false
dbhostport=localhost:27017
server.contextPath=/api
server.port=8080

#LDAP Server Url, including port of your LDAP server
#auth.ldapServerUrl=[ldap://company.com:389]
auth.ldapServerUrl=ldap://vds.xxx.com:389

#If using standard ldap
#LDAP User Dn Pattern, where the username is replaced with '{0}'
#auth.ldapUserDnPattern=[uid={0},OU=Users,dc=your,dc=company,dc=com]
auth.ldapUserDnPattern=uid={0},OU=people,dc=xxx,dc=com

API Console:

The entire log is below but I found this line which says Dn is empty. Is that is the reason, can you please let me know what the mistake is - [localhost-startStop-1] INFO org.springframework.security.ldap.DefaultSpringSecurityContextSource -  URL 'ldap://vds.xxx.com:389', root DN is ''

Complete Log:

[api_ldap.txt](https://github.com/capitalone/Hygieia/files/1042843/api_ldap.txt)

kavitha0624:
Hi Team,

Can anyone please help me with this issue

satishc1:
@kavitha0624 Are you using AD?

kavitha0624:
@satishc1 I did not get you.. what is AD?

kavitha0624:
Hi @satishc1 can you please let me know what did you mean by AD. This issue is blocking me. Please help.

satishc1:
@kavitha0624 Active Directory. The api properties for AD are different from the properties for LDAP, so if you are using AD, then make sure you set those.

kavitha0624:
@satishc1 , we are using LDAP here. So I have set the properties according to Hygieia setup document.

kavitha0624:
@satishc1 , could you please suggest how do I proceed with this configuration. As you asked, we are using LDAP here. So I am guessing the properties should be correct.

rob-miller-777:
[...]

PR #2320, opened 2018-06-25
Avoid Spring Security to query user's groups from LDAP

benoitm76:
### Quick resume
Replace DefaultLdapAuthoritiesPopulator by NullLdapAuthoritiesPopulator in order to:

- Avoid unnecessary LDAP Request
- Avoid `[LDAP: error code 32 - No Such Object]` in some case (#1439)

### In details

By default, Spring Security used DefaultLdapAuthoritiesPopulator in order to collect user's groups from LDAP.

If I understand well, **admin** role in Hygieia is assigned by the database and can not be assigned using LDAP groups (no parameter to configure Groups search in the LDAP and authorities retrieved using mongo database).

Due to DefaultLdapAuthoritiesPopulator, Spring Security does a LDAP request after login to get user's groups. I have replaced it by NullLdapAuthoritiesPopulator in order to remove this unnecessary request to LDAP.

Furthermore this can cause some LDAP errors regarding to the configuration of the server: `[LDAP: error code 32 - No Such Object]` (same problem than #1439)


gars01:
Great, it works

gars01:
Thanks a lot

rvema:
@benoitm76 We will review the changes and merge it .

benoitm76:
I have solved conflicts due to another merge.

benoitm76:
@rvema Unit test added ;-)

rvema:
We will review and merge it as soon as possible

PR #2339, opened 2018-07-05
Fix for performance api to pass in metrics

Sbrenthughes:
1) Fixes performance/create api not having the ability to pass in metrics.
2) Fixes CollectorType 
3) Fixes Null Pointer when hygieiaId isn't passed in


Sbrenthughes:
@tabladrum @nireeshT

PR #2366, opened 2018-07-20
[collectors/jenkins] properties-builder.sh: Fix MONGO_HOST parsing

abesto:


matthewbeasley:
@abesto - I know this pattern is weird but look more closely at the regex and you will see that what is in place correctly parses the mongo host and port.

abesto:
Ah, I totally missed the fact that `MONGO_PORT` should contain the protocol and host as well (even though it's in the comment above). Thank you.

PR #2367, opened 2018-07-20
Fix mongoDB script interpretation issue in Docker

dardanbekteshi:
I was getting JavaScript syntax errors when using the db-setup.js through piping it in the db-setup.sh shell script. I believe the reason is that the mongo interactive shell and the script execution use different interpreters, as described in the below documentation.

https://docs.mongodb.com/manual/tutorial/write-scripts-for-the-mongo-shell/#differences-between-interactive-and-scripted-mongo

Using the JavaScript equivalent for `use <database>` and executing the script instead of piping it seems to solve the aforementioned issue.


matthewbeasley:
@dardanbekteshi - I have not experienced this issue. Can you post a screenshot of what is failing for you and another screenshot of your fix working?

casz:
@dardanbekteshi is out of office atm however he will be back Wednesday 😅

matthewbeasley:
@dardanbekteshi - please respond to the previous comments.

PR #2427, opened 2018-09-06
Update pom.xml

riyas004:
frontend-maven-plugin  issue


rvema:
@riyas004 what was the reason behind upgrading the version of frontend-maven-plugin?

riyas004:
> 
> 
> @riyas004 what was the reason behind upgrading the version of frontend-maven-plugin?

When i install hygieia (mvn clean install package) to my system Finally getting UI error (maven frontend-plugin-0.0.29 issue, I'm just upgrade the plugin to 1.6 in inside the UI pom.xml. After I install (mvn clean install package), I didn't getting any error. That's the reason only i can update pom.xml.

rvema:
@riyas004 testing this PR locally and once successful will merge the same.

tabladrum:
@matthewbeasley can you check this out?

PR #2451, opened 2018-09-23
Corrected container image names

adityai:
Please ignore the .travis.yml changes. 

docker-compose.yml - the container image names were missing the prefix capitaloneio/ or capitalone/. I could not find jira and bitbucket collector images in capitalone's repo. So, I used the images made by liatrio.


adityai:
Corrected the merge issue from the previous pull request.

adityai:
Closed by mistake

rvema:
@adityai as per the changes we can't merge these changes as they seem to be specific to a certain use case. If there are specific changes to your environment please commit to your fork, Changes can be merged only if it suffices a larger audience. Hence I am closing this PR, please let me know if you have any concerns.

adityai:
When a user checks-out the code and runs 'docker-compose up -d', it fails because it cannot find the images on hub.docker.com. All of the images in the yml are missing a prefix like captialone and capitaloneio. In a couple of cases, the images aren't even available on hub.docker.com.

I am guessing that you have your own docker registry where you host all this stuff and it does not require a prefix like capitalone or capitaloneio. 

Unless these container names match whatever is out there on hub.docker.com, people out there like me will not be able to create docker images from the docker-compose.yml the way it is right now.

PR #2477, opened 2018-10-04
Add Ability to skip jobs from publishing data through Global Configuration

rvema:


hygieia-jenkins-plugin/src/main/java/jenkins/plugins/hygieia/HygieiaGlobalListener.java
```diff
@@ -34,58 +34,65 @@ public void onCompleted(Run run, @Nonnull TaskListener listener) {
...
+                                listener.getLogger().println("Hygieia: Auto Published " + gcir.getToolName() + " Data. " + genericItemResponse.toString());
+                            } else {
+                                listener.getLogger().println("Hygieia: Auto Published " + gcir.getToolName() + " Data. " + genericItemResponse.toString());
+                            }
+                        }
+                    } catch (IOException e) {
+                        listener.getLogger().println("Hygieia: Error Auto Publishing Generic Collector Item data." + '\n' + e.getMessage());
+                    }
+
+                }
```
tabladrum: Anything changed here or the diff is just due to format change?

rvema: just the format change, not sure why intellij did it that way.

PR #2506, opened 2018-10-17
Backward compatibility for POST endpoints. Jenkins plugin update: 

tabladrum:
## Changes
- Build, Code Quality, Test and Deploy POST end points now return document id and collector item id. This potentially breaks older clients. Added /v2/ rest endpoints for these keeping the old one as is. 

- Minor update to Jenkins plugin: If build publish fails, do not try to publish anything else as the build objectID is necessary to relate other information.


rvema:
@tabladrum looking into it now

PR #2513, opened 2018-10-18
API for the Github Webhook 

subodhbattina:
API for the Github Webhook: To process payloads for Pull Requests, Commits and Issues.


rvema:
review is in progress

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
```diff
@@ -0,0 +1,318 @@
...
+
+    @Override
+    public String process(JSONObject jsonObject) throws MalformedURLException, HygieiaException, ParseException {
+        String result = "Commits Processed Successfully";
+
+        String branch = null;
+        String repoUrl = null;
+
+        Object commitsObj = jsonObject.get("commits");
+        if ((commitsObj == null) || !(commitsObj instanceof List)) {
```
rvema: just checking instanceof should be sufficient here

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
```diff
@@ -0,0 +1,318 @@
...
+            commit.setScmRevisionNumber(commitId);
+
+            Object authorObjectFromCommit = cObj.get("author");
+            String authorName = restClient.getString(authorObjectFromCommit, "name");
+            commit.setScmAuthor(authorName);
+
+            String message = restClient.getString(cObj, "message");
+            commit.setScmCommitLog(message);
+
+            commit.setTimestamp(System.currentTimeMillis());
```
rvema: Shouldn't this be fetched from the payload? Consider the  case where there is an issue with Github and the payloads to webhooks are delivered late.

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
```diff
@@ -0,0 +1,318 @@
...
+            commit.setScmUrl(repoUrl);
+            commit.setScmBranch(branch);
+            commit.setScmAuthorLDAPDN(authorLDAPDN);
+
+            DateTime commitTimestamp = new DateTime(restClient.getString(cObj, "timestamp"));
+            DateTime commitTimestampStepBack = commitTimestamp.minusMinutes(gitHubWebHookSettings.getCommitTimestampOffset());
+
+            commit.setScmCommitTimestamp(commitTimestamp.getMillis());
```
rvema: Shouldn't this be fetched from payload? Consider a scenario where github sends the payload late due to issues.

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
```diff
@@ -0,0 +1,318 @@
...
+                                            .collect(Collectors.toList());
+
+        // In case of Rebase and Merge, only the last commit in the list of commits on the "merge commit" json should have the PR number.
+        // This is because, there should be a corresponding PR in the DB whose "merge_commit_sha" matches the last commit in the list of commits
+        if (!CollectionUtils.isEmpty(commitsWithPullNumber)
+                && (commitsWithPullNumber.size() == 1)
+                && (commitsList.size() > 1)) {
+            Commit commitWithPR = commitsWithPullNumber.get(0);
+
+            Optional.ofNullable(commitsList)
```
rvema: Optional.ofNullable may not be required as as you are checking for size() in line 204

subodhbattina: Good Catch, thanks. I have cleaned up the null checks before "instanceof" usage.

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
```diff
@@ -0,0 +1,318 @@
...
+        if (target == null) { return null; }
+
+        JSONObject history = (JSONObject) target.get("history");
+        if (history == null) { return null; }
+
+        JSONArray edges = (JSONArray) history.get("edges");
+        if (CollectionUtils.isEmpty(edges)) { return null; }
+
+        for (Object o : edges) {
+            JSONObject node = (JSONObject) ((JSONObject) o).get("node");
```
rvema: check for instanceof just to be safe

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
subodhbattina: Have taken care of this

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
subodhbattina: Have removed the Optional usage.

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubV3.java
```diff
@@ -0,0 +1,201 @@
...
+import java.net.MalformedURLException;
+import java.util.HashMap;
+import java.util.Map;
+
+public abstract class GitHubV3 {
+    private static final Log LOG = LogFactory.getLog(GitHubV3.class);
+
+    private static final String REPO_URL = "url";
+    private static final String BRANCH = "branch";
+    protected static final String USER_ID = "userID";
```
rvema: any reason why these need to be protected? they can be marked private

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
```diff
@@ -0,0 +1,318 @@
...
+            }
+
+            Commit commit = new Commit();
+
+            String commitId = restClient.getString(cObj, "id");
+            commit.setScmRevisionNumber(commitId);
+
+            Object authorObjectFromCommit = restClient.getAsObject(cObj,"author");
+            String authorName = restClient.getString(authorObjectFromCommit, "name");
+            commit.setScmAuthor(authorName);
```
rvema: variables authorName, message, commitId can be avoided and directly set into the instance variable commit as not other operation is done on them

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
```diff
@@ -0,0 +1,318 @@
...
+        List<Commit> commitsWithPullNumber = Optional.ofNullable(commitsList)
+                                            .orElseGet(Collections::emptyList).stream()
+                                            .filter(commit -> !StringUtils.isEmpty(commit.getPullNumber()))
+                                            .collect(Collectors.toList());
+
+        // In case of Rebase and Merge, only the last commit in the list of commits on the "merge commit" json should have the PR number.
+        // This is because, there should be a corresponding PR in the DB whose "merge_commit_sha" matches the last commit in the list of commits
+        if (!CollectionUtils.isEmpty(commitsWithPullNumber)
+                && (commitsWithPullNumber.size() == 1)
+                && (commitsList.size() > 1)) {
```
rvema: There is a possiblity of NPE for commitsList

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
```diff
@@ -0,0 +1,318 @@
...
+        return result;
+    }
+
+    protected List<Commit> getCommits(List<Map> commitsObjectList, String repoUrl,
+                                      String branch, String authorLDAPDN) throws MalformedURLException, HygieiaException, ParseException {
+        List<Commit> commitsList = new ArrayList<>();
+
+        GitHubParsed gitHubParsed = new GitHubParsed(repoUrl);
+        CollectorItem collectorItem = getCollectorItem(gitHubParsed.getUrl(), branch);
+        GitHubWebHookSettings gitHubWebHookSettings = parseAsGitHubWebHook(apiSettings.getGitHubWebHook());
```
rvema: handle null checks for gitHubWebHookSettings there is a possibility parseAsGitHubWebHook() can return it as null

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
```diff
@@ -0,0 +1,318 @@
...
+        GitRequest pr = gitRequestRepository.findByScmRevisionNumberOrScmMergeEventRevisionNumber(commit.getScmRevisionNumber());
+        if (pr == null) {
+            pr = gitRequestRepository.findByCommitScmRevisionNumber(commit.getScmRevisionNumber());
+        }
+        if (pr != null) {
+            commit.setPullNumber(pr.getNumber());
+        }
+    }
+
+    protected void setCollectorItemId (Commit commit, CollectorItem collectorItem) {
```
rvema: Can this method be refactored such that it can be reused by all 3 functionalities - commits, PRs and Issues?

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubIssueV3.java
```diff
@@ -0,0 +1,138 @@
...
+            issue.setResolutiontime((closedTimestamp - createdTimestamp));
+            issue.setState("closed");
+        }
+
+        setCollectorItemId(issue);
+
+        return issue;
+    }
+
+    protected void setCollectorItemId(GitRequest issue) {
```
rvema: see comment on GitHubCommitV3 for this method

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubPullRequestV3.java
```diff
@@ -0,0 +1,503 @@
...
+        if (gitHubWebHookSettings != null) {
+            gitHubWebHookToken = gitHubWebHookSettings.getToken();
+        }
+
+        long start = System.currentTimeMillis();
+
+        String repoToken = getRepositoryToken(repoUrl);
+
+        long end = System.currentTimeMillis();
+        LOG.debug("Time to make gitHubRepoRepository call = "+(end-start));
```
rvema: Can you change the verbiage to say gitHubRepoRepository call to fetch the repo token

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubPullRequestV3.java
```diff
@@ -0,0 +1,503 @@
...
+        if (existingPR != null) {
+            pull.setId(existingPR.getId());
+            pull.setCollectorItemId(existingPR.getCollectorItemId());
+        }
+
+        gitRequestRepository.save(pull);
+
+        long end = System.currentTimeMillis();
+
+        LOG.debug("Time to make gitRequestRepository call = "+(end-start));
```
rvema: please change the verbiage to be more meaningful

core/src/main/java/com/capitalone/dashboard/client/RestClient.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    protected HttpHeaders createHeaders(String key, String token) {
+        String authHeader = key.trim() + " " + token.trim();
+        HttpHeaders headers = new HttpHeaders();
+        headers.set("Authorization", authHeader);
+        return headers;
+    }
+
+    public JSONObject parseAsObject(ResponseEntity<String> response) throws ParseException {
```
rvema: all the standalone utility methods that do not depend on restoperations should be static methods and if possible move it to a separate utility class.

core/src/main/java/com/capitalone/dashboard/custom/GitHubParsed.java
```diff
@@ -0,0 +1,80 @@
+package com.capitalone.dashboard.custom;
```
rvema: Any reason why this class is not part of the webhook package series?

rvema:
Also please include the security configuration for the new controller (/webhook/github/v3) else the calls to api will fail from github.

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubV3.java
subodhbattina: This was by oversight, have made them private, thanks

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
subodhbattina: Actually, from the caller of this function perspective, there is no NPE possibility in this case. But anyways, I added the check, thanks.

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
subodhbattina: > variables authorName, message, commitId can be avoided and directly set into the instance variable commit as not other operation is done on them

Cleaned up the variable "authorName", "message" and "commitId" are being used else where, thanks !

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
subodhbattina: Good Catch. I have put this check in, thanks.

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubPullRequestV3.java
subodhbattina: Ok, I have updated this, thanks.

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubPullRequestV3.java
subodhbattina: Ok, have updated this, thanks.

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubV3.java
```diff
@@ -0,0 +1,201 @@
...
+        long start = System.currentTimeMillis();
+
+        QCollectorItem item = new QCollectorItem("collectorItem");
+        BooleanBuilder builder = new BooleanBuilder();
+        builder.and(item.collectorId.eq(collectorId));
+        options.keySet().forEach(k -> {
+            builder.and(item.options.get(k).eq(options.get(k)));
+        });
+
+        Iterable<GitHubRepo> gitHubRepoIterable = getGitHubRepoRepository().findAll(builder.getValue());
```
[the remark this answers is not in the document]
subodhbattina: Good Catch, by oversight, missed to put the while loop : ). Have fixed it, thanks.

core/src/main/java/com/capitalone/dashboard/client/RestClient.java
subodhbattina: While in theory it sounds good, but I am afraid practically it presents challenges when building our unit tests around static methods. Mockito doesn't work very well with static methods, thanks

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubCommitV3.java
subodhbattina: I see what you are saying, but since each of the setCollectorItemId methods, although they follow a similar pattern, use different object types and repository types. So I think even if I refactored them to take care of the three cases (commits, PRs and Issues), its not going to be pretty. The method would be one big monolithic method with a bunch of if-else conditions ..

api/src/main/java/com/capitalone/dashboard/webhook/github/GitHubIssueV3.java
subodhbattina: Same reply thoughts as on GitHubCommitV3, thanks

subodhbattina:
> Also please include the security configuration for the new controller (/webhook/github/v3) else the calls to api will fail from github.

Reply: I have incorporated an authentication filter for the github webhook api, thanks

core/src/main/java/com/capitalone/dashboard/custom/GitHubParsed.java
subodhbattina: I have moved GitHubParsed into the same package as the other models for github webhook api, thanks

PR #2534, opened 2018-10-26
Delete all test bug

kheinz04:
PR for issue TEART-21795

When a field is changed to empty in the quality widget it was not able to be saved, as the associateCollectorToComponent function only updates based off of new values being passed, not values omitted. This fix checks to see if other values from the quality widget are being updated so that if a value is missing, it can be assumed that the previous value (if there was any) was deleted.

**Caveat: this will not work if all values from the widget are deleted from the configure page as there is no way to determine the widget type is there are no collector types in the request. We would need to keep track of this relationship in order to fix this functionality. As an alternative, the delete button will delete the entire widget and all values, resulting in the same effect.


api/src/main/java/com/capitalone/dashboard/service/DashboardServiceImpl.java
```diff
@@ -46,6 +46,8 @@
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
 
+import com.capitalone.dashboard.model.CollectorType.*;
```
rvema: please avoid importing using *, import only the classes that are required.

api/src/main/java/com/capitalone/dashboard/service/DashboardServiceImpl.java
```diff
@@ -271,6 +273,24 @@ public Component associateCollectorToComponent(ObjectId componentId, List<Object
                 }
                 // remove all collector items of a type
                 component.getCollectorItems().remove(collector.getCollectorType());
+
+
+            }
+        }
+
+        // If a collector type is within the code analysis widget, check to see if any of the remaining fields were passed values
+        if(incomingTypes.contains(CollectorType.Test) || incomingTypes.contains(CollectorType.StaticSecurityScan) || incomingTypes.contains(CollectorType.CodeQuality) || incomingTypes.contains(CollectorType.LibraryPolicy) ){
```
rvema: this if condition seems redundant. can you revisit this?

PR #2555, opened 2018-11-05
Added feature toggle to lookup dashboards

rvema:
Added a feature toggle via ApiSettings to lookup dashboards for BuildData create request(compatible from v3 onwards)

### DIFF
--- a/api/src/main/java/com/capitalone/dashboard/service/BuildServiceImpl.java
+++ b/api/src/main/java/com/capitalone/dashboard/service/BuildServiceImpl.java
@@ -23,7 +23,6 @@ import com.capitalone.dashboard.response.BuildDataCreateResponse;
 import com.mysema.query.BooleanBuilder;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang3.BooleanUtils;
 import org.joda.time.LocalDate;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.data.domain.PageRequest;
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/capitalone/dashboard/service/BuildServiceImpl.java
```diff
@@ -162,7 +168,9 @@ public BuildDataCreateResponse createV3(BuildDataCreateRequest request) throws H
             throw new HygieiaException(e);
         }
         finally {
-            populateDashboardId(response);
+            if(BooleanUtils.isTrue(settings.isLookupDashboardForBuildDataCreate())) {
```
nireeshT: @rvema  remove BooleanUtils and use implicit condition checking 
if(settings.isLookupDashboardForBuildDataCreate()))

api/src/test/java/com/capitalone/dashboard/service/BuildServiceTest.java
```diff
@@ -55,6 +52,8 @@
     @Mock private DashboardServiceImpl dashboardService;
     @Mock
     private CollectorService collectorService;
+    @Mock
+    private ApiSettings apiSettings;
```
nireeshT: @rvema  are we using this Mock ?

api/src/main/java/com/capitalone/dashboard/service/BuildServiceImpl.java
rvema: Booleanutils is null safe hence it was used , any specific  reason behind removing it?

api/src/test/java/com/capitalone/dashboard/service/BuildServiceTest.java
rvema: Yes the when you mock the class it provides the mocked version of the same during execution and hence it was used

rvema: added additional testcase when the lookupDashboardForBuildDataCreate is configured to return true

api/src/main/java/com/capitalone/dashboard/service/BuildServiceImpl.java
rvema: addressed it.

PR #2579, opened 2018-11-15
updating core version to 3.0.6 for all projects

kumarvarun1252:
Updated the core version to 3.0.6 for all the projects


rvema:
This is no longer required as we fixed it part of other code changes.

PR #2580, opened 2018-11-15
Jira Features Widget creation via remote API

subodhsbattina:
Updates to enable the user to create a jira feature widget using remoteCreate API


Sbrenthughes:
This needs to be broken up into 2 PRs.

api/src/main/java/com/capitalone/dashboard/request/DashboardRemoteRequest.java
```diff
@@ -248,7 +245,34 @@ public String getWidgetName() {
...
+            Map<String, Object> opts = new HashMap<>();
+            opts.put("id", "feature0");
+            options.keySet().forEach(key -> {
+                opts.put(key, options.get(key));
+            });
+            return opts;
+        }
+
+        @Override
+        public CollectorItem toCollectorItem(Collector collector) throws HygieiaException {
```
Sbrenthughes: Can you test without overwriting the collectorItem? As long as you pass the correct fields it should work.

subodhbattina: Hi @Sbrenthughes Yes, it does work without overwriting this method, I have tested that flow as well. The intent to overwrite this method was to create collector item with only unique options as oppose to all the options, as is the case when the widget is created via the UI. I am ok either ways, just want to do what would be correct in this case, I will circle back on this soon. Thanks

PR #2612, opened 2018-12-03
Opensource icon fix

kheinz04:
Fixed the icon to be related to the top violation type of a specific violation type in order to match the displayed value. 

https://jira.kdc.capitalone.com/browse/TEART-23987


kheinz04:
@Sbrenthughes Properties file was removed

PR #2635, opened 2018-12-13
Opensource redesign

kheinz04:
Updated the opensource widget to list a summary of all of the violation types as opposed to just the highest group. 

Added a 'total' and a 'view all' option to link to the details of scan.

Hid Nexus information if it is not being used (if there is no data)


kheinz04:
@nireeshT - PR based on what we reviewed today

kheinz04:
image 4
image 3

PR #2663, opened 2019-01-09
Docker container fails to run without dbhostport property.

vishalraina:
Docker container for api server fails to run if you run DB outside docker compose environment,
By Adding dbhostport property this issue is resolved.


rvema:
closing this PR since it has user specific information and this file needs to be modified per user's environment

PR #2706, opened 2019-02-01
Quality widget loop base version

kheinz04:
Update Quality widget to switch between tabs on a loop 
image
image

Updated logo:
image
image


UI/src/components/templates/capone.js
```diff
@@ -9,8 +9,8 @@
         .module(HygieiaConfig.module)
         .controller('CapOneTemplateController', CapOneTemplateController);
 
-    CapOneTemplateController.$inject = [];
-    function CapOneTemplateController() {
+    CapOneTemplateController.$inject = ['$interval'];
```
nireeshT: @kheinz04 do we need to inject $interval here

UI/src/components/templates/capone.html
```diff
@@ -48,7 +48,9 @@ <h4 title="{{::ctrl.dashboard.title }}">{{::ctrl.dashboard.title | trunText:50}}
                         </li>
```
nireeshT: @kheinz04 this file has no changes , do not commit

UI/src/components/templates/capone.js
kheinz04: Good catch, that was from another update when I was trying to loop the performance tab as well. I will remove.

PR #2722, opened 2019-02-08
Adding in support for multiple owners 

andrewalvintran:
Dependency on this: https://github.com/Hygieia/hygieia-core/pull/45

Has been tested manually, but should have more test cases.  Right now I see user.json is being used to load into userInfoRepository for DashboardRemoteServiceTest.  Need to figure out the best approach for adding more test cases.


Sbrenthughes:
@andrewalvintran Can you confirm that a user cannot be added to a dashboard more than once?

andrewalvintran:
@Sbrenthughes Yes, I wrote a test for that, where the owner is repeated in the owners field and the same owner shows up multiple times in the owners field.

api/src/main/java/com/capitalone/dashboard/service/DashboardRemoteServiceImpl.java
```diff
@@ -56,18 +59,51 @@ public DashboardRemoteServiceImpl(
...
+        List<Owner> owners = metaData.getOwners();
+
+        if (owner == null && owners == null) {
+            throw new HygieiaException("There are no owner/owners field in the request", HygieiaException.INVALID_CONFIGURATION);
+        }
+
+        if (owners == null) {
+            owners = new ArrayList<Owner>();
+            owners.add(owner);
+        } else if (owner != null) {
```
rvema: This condition is unnecessary

api/src/main/java/com/capitalone/dashboard/request/DashboardRequest.java
```diff
@@ -130,11 +131,13 @@ public Dashboard toDashboard() {
         DashboardType type = DashboardType.fromString(this.type);
         Application application = new Application(applicationName, new Component(componentName));
         Owner owner = new Owner(AuthenticationUtil.getUsernameFromContext(), AuthenticationUtil.getAuthTypeFromContext());
+        List<Owner> owners = new ArrayList<Owner>();
```
rvema: You could use Lists.newArrayList()

api/src/main/java/com/capitalone/dashboard/service/DashboardRemoteServiceImpl.java
```diff
@@ -56,18 +59,51 @@ public DashboardRemoteServiceImpl(
...
+     * @param request
+     * @return List<Owner> list of owners to be added to the dashboard
+     * @throws HygieiaException
+     */
+    private List<Owner> getOwners(DashboardRemoteRequest request) throws HygieiaException {
+        DashboardRemoteRequest.DashboardMetaData metaData = request.getMetaData();
+        Owner owner = metaData.getOwner();
+        List<Owner> owners = metaData.getOwners();
+
+        if (owner == null && owners == null) {
```
rvema: shouldn't you be checking if the list of owners is not empty? It should be changed to (owner==null && CollectionUtils.isEmpty(owners))

api/src/main/java/com/capitalone/dashboard/service/DashboardRemoteServiceImpl.java
```diff
@@ -56,18 +59,51 @@ public DashboardRemoteServiceImpl(
...
     public Dashboard remoteCreate(DashboardRemoteRequest request, boolean isUpdate) throws HygieiaException {
         Dashboard dashboard;
         Map<String, Widget> existingWidgets = new HashMap<>();
 
-        if (!userInfoService.isUserValid(request.getMetaData().getOwner().getUsername(), request.getMetaData().getOwner().getAuthType())) {
-            throw new HygieiaException("Invalid owner information or authentication type. Owner first needs to sign up in Hygieia", HygieiaException.BAD_DATA);
+        List<Owner> owners = getOwners(request);
+
+        for (Owner owner : owners) {
+            if (!userInfoService.isUserValid(owner.getUsername(), owner.getAuthType())) {
```
rvema: Do we want to fail even if one of the owners passed in is bad?

api/src/main/java/com/capitalone/dashboard/service/DashboardRemoteServiceImpl.java
```diff
@@ -56,18 +59,51 @@ public DashboardRemoteServiceImpl(
...
+                throw new HygieiaException("Invalid owner information or authentication type. Owner first needs to sign up in Hygieia", HygieiaException.BAD_DATA);
+            }
         }
 
         List<Dashboard> dashboards = findExistingDashboardsFromRequest( request );
         if (!CollectionUtils.isEmpty(dashboards)) {
             dashboard = dashboards.get(0);
+            Set<Owner> uniqueOwners = new HashSet<Owner>(owners);
+            uniqueOwners.addAll(dashboard.getOwners());
```
rvema: You could use Collections.list() here

api/src/main/java/com/capitalone/dashboard/request/DashboardRemoteRequest.java
```diff
@@ -68,9 +68,10 @@
         @NotNull
         private String componentName;
 
-        @NotNull
         Owner owner;
```
rvema: Why are we still having single owner when you have the list of owners?

api/src/main/java/com/capitalone/dashboard/service/DashboardServiceImpl.java
```diff
@@ -441,10 +444,10 @@ public boolean apply(Collector collector) {
     public Iterable<Owner> updateOwners(ObjectId dashboardId, Iterable<Owner> owners) {
         for(Owner owner : owners) {
         	String username = owner.getUsername();
-        	AuthType authType = owner.getAuthType();
-        	if(userInfoRepository.findByUsernameAndAuthType(username, authType) == null) {
-        		throw new UserNotFoundException(username, authType);
-        	}
+            AuthType authType = owner.getAuthType();
+            if (!userInfoService.isUserValid(username, authType)) {
```
rvema: check if we want to fail even if one of the users passed in is bad.

api/src/main/java/com/capitalone/dashboard/request/DashboardRemoteRequest.java
rvema: This field can stay to support backward compatibility

api/src/main/java/com/capitalone/dashboard/service/DashboardRemoteServiceImpl.java
andrewalvintran: Since we don't denote @NotNull for DashboardMetaData, don't we have to check here if one is null?  If someone only passes in owner, the owners field will be null, and likewise for passing in owners, the owner field will be null.  I tried removing this condition but it fails with "internal error" if only one of the fields is given.

api/src/main/java/com/capitalone/dashboard/service/DashboardRemoteServiceImpl.java
andrewalvintran: I would assume no, but that's up to us to decide, should I change it to instead remove the owner from the list?

api/src/main/java/com/capitalone/dashboard/service/DashboardRemoteServiceImpl.java
andrewalvintran: So in the getOwners function, I validate the current owner/owners list to make sure there are no duplicates, but this is not done for the existing owners.  From my understanding, Collections.list() returns an ArrayList, but I need the Set to make sure that there are no duplicates.  Is there another way you want me to use it?

api/src/main/java/com/capitalone/dashboard/service/DashboardRemoteServiceImpl.java
rvema: just skip if one of the owner is bad and proceed to the next owner passed in the list.

api/src/main/java/com/capitalone/dashboard/service/DashboardRemoteServiceImpl.java
rvema: The previous comment was for line 106 which can be written as dashboard.setOwners(Collections.list(uniqueOwners));

PR #2761, opened 2019-03-01
Test result audit response update

nameisaravind:
This PR is created to add feature test result counts and type in test result audit response


nameisaravind:
> Can you add some test cases for this?

tests added

PR #2778, opened 2019-03-07
enabled work github under proxy server

poojithkumar:
made github collectors work under proxy server
issue solved: collector is up but not get data


poojithkumar:
hi

Sbrenthughes:
My suggestion would be to do this: https://github.com/Hygieia/Hygieia/blob/ae661508aeb66b1aa0b022e90f69d2a59958cdce/collectors/feature/jira/src/main/java/com/capitalone/dashboard/collector/FeatureCollectorTask.java#L124

and then add:
`http.proxyUser` and `https.proxyPassword`

poojithkumar:
I will try that ..let you know about pros and cons.

Sbrenthughes:
Sure my only thought is this only updates 2 classes. The settings class and the CollectorTask class.

poojithkumar:
Ok.. thanks for the valuable advice.

poojithkumar:
@Sbrenthughes  made changes according to your advice.
one small question.
which technology you are using in Hygieia ? (are you using grafana ? )

collectors/scm/github/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
```diff
@@ -32,6 +32,7 @@
 import org.springframework.web.client.RestClientException;
 import org.springframework.web.client.RestOperations;
 
+
```
Sbrenthughes: Can you remove all of the formatting? There are unnecessary changes and can make things harder to read.

collectors/scm/github/src/main/java/com/capitalone/dashboard/collector/GitHubCollectorTask.java
```diff
@@ -166,9 +166,20 @@ public void collect(Collector collector) {
...
-
+       
         clean(collector);
-        for (GitHubRepo repo : enabledRepos(collector)) {
+        
+        String proxyUrl = gitHubSettings.getProxy();
+        String proxyPort = gitHubSettings.getProxyPort();
+ 
+        if (!StringUtils.isEmpty(proxyUrl) && !StringUtils.isEmpty(proxyPort)) {
+            System.setProperty("http.proxyHost", proxyUrl);
```
Sbrenthughes: Nice work here. One more thing can you add two more property options? `proxyUser` and `proxyPassword ` here and place it in a StringUtils.isEmpty check? Similar to Line 175. Some users may need to provide authentication.

collectors/scm/github/src/main/java/com/capitalone/dashboard/collector/GitHubSettings.java
```diff
@@ -12,50 +12,65 @@
 @Component
 @ConfigurationProperties(prefix = "github")
 public class GitHubSettings {
+    private String proxyPort;
+    private String proxy;
     private String cron;
     private String host;
     private String key;
-	@Value("${github.firstRunHistoryDays:14}")
+        @Value("${github.firstRunHistoryDays:14}")
```
Sbrenthughes: All of this formatting needs to be reset.

poojithkumar:
above code  makes github collector to work under proxy server
add proxy details in  githubs properties file
example : add as below

github.proxy=<your-proxyUrl>
github.proxyPort=<your-proxyPort>
github.proxyUser=< your- authentication-userName>
github.proxyPassword=<your-authentication- Password>

thank you

collectors/scm/github/src/main/java/com/capitalone/dashboard/collector/GitHubCollectorTask.java
```diff
@@ -166,9 +166,28 @@ public void collect(Collector collector) {
...
+        String proxyUser= gitHubSettings.getProxyUser();
+        String proxyPassword= gitHubSettings.getProxyPassword();
+        
+        if (!StringUtils.isEmpty(proxyUrl) && !StringUtils.isEmpty(proxyPort)) {
+            System.setProperty("http.proxyHost", proxyUrl);
+            System.setProperty("https.proxyHost", proxyUrl);
+            System.setProperty("http.proxyPort", proxyPort);
+            System.setProperty("https.proxyPort", proxyPort);
+        }
+         if (!StringUtils.isEmpty(proxyUser) && !StringUtils.isEmpty(proxyPassword)) {
```
Sbrenthughes: This if block would need to be in the above if block

poojithkumar: i don't think that change is required. that if (183) inside other if statement make sense. because username and password is valid only if proxy url is given . without url and port no .user id and password make no sense

Sbrenthughes: I’d prefer for this to be inside the if block on 177. This is because if the Url and port doesn’t exist why even check to see if the user and password exist?

collectors/scm/github/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
```diff
@@ -721,4 +721,4 @@ private static String getTimeForApi (Date dt) {
     }
 }
 
-// X-RateLimit-Remaining
```
Sbrenthughes: Revert this file change

collectors/scm/github/src/main/java/com/capitalone/dashboard/collector/GitHubCollectorTask.java
```diff
@@ -166,9 +166,29 @@ public void collect(Collector collector) {
...
+            System.setProperty("https.proxyHost", proxyUrl);
+            System.setProperty("http.proxyPort", proxyPort);
+            System.setProperty("https.proxyPort", proxyPort);
+        
+            if (!StringUtils.isEmpty(proxyUser) && !StringUtils.isEmpty(proxyPassword)) {
+                System.setProperty("http.proxyUser", proxyUser);
+                System.setProperty("https.proxyUser", proxyUser);
+                System.setProperty("http.proxyPassword", proxyPassword);
+                System.setProperty("https.proxyPassword", proxyPassword);
+        }
```
Sbrenthughes: Fix this formatting

collectors/scm/github/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
poojithkumar: @Sbrenthughes whats wrong with that file. i don't see any difference over there.

PR #2781, opened 2019-03-07
Resolves Travis CI mvn deploy of non-snapshot 

shriver135:
The lack of snapshot causes the authorization error we are seeing, minor version bump to avoid conflicts with previous 3.0.1 release, deploy snapshot updated to build the same as full release,  full release mimicing the removal of snapshot before deploy command.


UI-tests/pom.xml
```diff
@@ -7,7 +7,7 @@
 	<parent>
 		<groupId>com.capitalone.dashboard</groupId>
 		<artifactId>Hygieia</artifactId>
-		<version>3.0.1</version>
+		<version>3.0.2-SNAPSHOT</version>
```
nireeshT: @shriver135  can you create property for version in Hygieia/pom.xml and use it across.

in  <properties>
...
</properties>

PR #2826, opened 2019-04-10
Graphql collector fix for redirected repos

andrewalvintran:
When an error is thrown for a GraphQL query, the errors are returned in the JSON response.  The errors are checked for "NOT_FOUND", which means that the repo could have been redirected.  A REST call is made to the original repo to look for the new repo url.

### DIFF
--- a/collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
+++ b/collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
@@ -231,7 +231,7 @@ public class DefaultGitHubClient implements GitHubClient {
         connectCommitToPulls();
     }
 
-    private RedirectedStatus checkForRedirectedRepo(GitHubRepo repo) throws MalformedURLException, HygieiaException {
+    public RedirectedStatus checkForRedirectedRepo(GitHubRepo repo) throws MalformedURLException, HygieiaException {
         GitHubParsed gitHubParsed = new GitHubParsed(repo.getRepoUrl());
         String query = gitHubParsed.getBaseApiUrl() + "repos/" + gitHubParsed.getOrgName() + "/" + gitHubParsed.getRepoName();
 
@@ -241,10 +241,9 @@ public class DefaultGitHubClient implements GitHubClient {
         JSONParser parser = new JSONParser();
         try {
             JSONObject queryJSONBody = (JSONObject) parser.parse(queryBody);
-
             String repoUrl = (String) queryJSONBody.get("html_url");
-            LOG.debug("original url: " + repo.getRepoUrl() + " is redirected to new url: " + repoUrl);
             if (!repoUrl.equals(repo.getRepoUrl())) {
+                LOG.debug("original url: " + repo.getRepoUrl() + " is redirected to new url: " + repoUrl);
                 return new RedirectedStatus(true, repoUrl);
             }
 
--- a/collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
+++ b/collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
@@ -71,6 +71,26 @@ public class DefaultGitHubClient implements GitHubClient {
 
     private static final int FIRST_RUN_HISTORY_DEFAULT = 14;
 
+    public class RedirectedStatus {
+        private boolean isRedirected = false;
+        private String redirectedUrl = null;
+
+        RedirectedStatus() {}
+
+        RedirectedStatus(boolean isRedirected, String redirectedUrl) {
+            this.isRedirected = isRedirected;
+            this.redirectedUrl = redirectedUrl;
+        }
+
+        String getRedirectedUrl() {
+            return this.redirectedUrl;
+        }
+
+        boolean isRedirected() {
+            return this.isRedirected;
+        }
+    }
+
     @Autowired
     public DefaultGitHubClient(GitHubSettings settings,
                                Supplier<RestOperations> restOperationsSupplier) {
@@ -141,13 +161,8 @@ public class DefaultGitHubClient implements GitHubClient {
         int loopCount = 1;
         while (!alldone) {
             LOG.debug("Executing loop " + loopCount + " for " + gitHubParsed.getOrgName() + "/" + gitHubParsed.getRepoName());
-            ResponseEntity<String> response = makeRestCallPost(graphQLurl, repo.getUserId(), decryptedPassword, decryptPersonalAccessToken, query);
-            JSONObject data = (JSONObject) parseAsObject(response).get("data");
-            JSONArray errors = getArray(parseAsObject(response), "errors");
+            JSONObject data = getDataFromRestCallPost(graphQLurl, repo, decryptedPassword, decryptPersonalAccessToken, query);
 
-            if (!CollectionUtils.isEmpty(errors)) {
-                throw new HygieiaException("Error in GraphQL query:" + errors.toJSONString(), HygieiaException.JSON_FORMAT_ERROR);
-            }
             if (data != null) {
                 JSONObject repository = (JSONObject) data.get("repository");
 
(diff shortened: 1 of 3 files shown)


tabladrum:
I think, we should update the corresponding collector item with the "new" home of the repo so that going forward, the first try will be successful and also all other associated data in Hygieia will be clean.

tabladrum:
@andrewalvintran @nireeshT this was a force push - please review together to make sure things are still good.

andrewalvintran:
The force push was just a rebase with my branch in it.  I will review with @nireeshT

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
```diff
@@ -1042,6 +1056,42 @@ private static DateTime getDate(DateTime dateInstance, int offsetDays, int offse
...
+
+        if (!CollectionUtils.isEmpty(errors)) {
+            JSONObject error = (JSONObject) errors.get(0);
+            if (error.get("type").equals("NOT_FOUND")) {
+                RedirectedStatus redirectedStatus = checkForRedirectedRepo(repo);
+
+                if (redirectedStatus.isRedirected()) {
+                    String redirectedUrl = redirectedStatus.getRedirectedUrl();
+                    LOG.debug("Repo was redirected from: " + repo.getRepoUrl() + " to " + redirectedUrl);
+                    repo.setRepoUrl(redirectedUrl);
```
andrewalvintran: > I think, we should update the corresponding collector item with the "new" home of the repo so that going forward, the first try will be successful and also all other associated data in Hygieia will be clean.

Not sure how the collector items are associated to the repos and I will review this with Nireesh, but I update the repo url here to point to the redirected repo and each run after that only looks at the "new" home of the repo.

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
```diff
@@ -223,8 +230,23 @@ public void fireGraphQL(GitHubRepo repo, boolean firstRun, Map<Long, String> exi
...
+    public RedirectedStatus checkForRedirectedRepo(GitHubRepo repo) throws MalformedURLException, HygieiaException {
+        GitHubParsed gitHubParsed = new GitHubParsed(repo.getRepoUrl());
+        String query = gitHubParsed.getBaseApiUrl() + "repos/" + gitHubParsed.getOrgName() + "/" + gitHubParsed.getRepoName();
+
+        ResponseEntity<String> response = makeRestCallGet(query);
+
+        JSONObject queryJSONBody = (JSONObject) parseAsObject(response);
+        String repoUrl = (String) queryJSONBody.get("html_url");
+        if (!repoUrl.equals(repo.getRepoUrl())) {
+            LOG.debug("original url: " + repo.getRepoUrl() + " is redirected to new url: " + repoUrl);
```
benj58xu: To be splunk/ELK friendly, I would use LOG.info instead of LOG.debug, and use the format "URL redirection: original=<url>, redirected=<url>"

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
benj58xu: Redirection sometimes is used as a normal practice, to hide URLs that are often changes, so not sure if it is a good idea to change to the redirected URL permanently. We may want to simply log the redirections, and manually confirm whether the change is permanent or not.

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
```diff
@@ -1042,6 +1056,42 @@ private static DateTime getDate(DateTime dateInstance, int offsetDays, int offse
...
+                        variableJSON.put("name", gitHubParsed.getRepoName());
+                        variableJSON.put("owner", gitHubParsed.getOrgName());
+                        query.put("variables", variableJSON.toString());
+                    } catch (ParseException e) {
+                        LOG.error("Could not parse JSON String", e);
+                    }
+                    return getDataFromRestCallPost(url, repo, password, personalAccessToken, query);
+                }
+            }
+
```
benj58xu: The code handles one level of redirection. Not sure we need to handle 2 or more levels of redirections?

andrewalvintran: As far as I can tell from trying it out with test repos, all of the repos will get redirected to the end repo, so there should only be one level.

For example if I renamed a repo from org/repo1 -> org/repo2 -> org/repo3, the org/repo1 will redirect directly to org/repo3 in the `html_url` section of the response body.

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
Sbrenthughes: @tabladrum @andrewalvintran The collectorItem appears to be updated. However, the component is not be updated to reflect that change.

Sbrenthughes: @benj58xu I can't see an instance where a team would change orgs or repo names often.

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
Sbrenthughes: @andrewalvintran does  org/repo2  also point to  org/repo3?

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
```diff
@@ -1042,6 +1056,42 @@ private static DateTime getDate(DateTime dateInstance, int offsetDays, int offse
         return dateInstance.minusDays(offsetDays).minusMinutes(offsetMinutes);
     }
 
+
+    private JSONObject getDataFromRestCallPost(String url, GitHubRepo repo, String password, String personalAccessToken, JSONObject query) throws MalformedURLException, HygieiaException {
+        ResponseEntity<String> response = makeRestCallPost(url, repo.getUserId(), password, personalAccessToken, query);
+        JSONObject data = (JSONObject) parseAsObject(response).get("data");
+        JSONArray errors = getArray(parseAsObject(response), "errors");
+
+        if (!CollectionUtils.isEmpty(errors)) {
```
Sbrenthughes: My only comment here is that this file is riddled with deeply nested ifs and loops. Could you update this to fail fast to remove some of nests?

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
```diff
@@ -223,8 +230,23 @@ public void fireGraphQL(GitHubRepo repo, boolean firstRun, Map<Long, String> exi
...
     }
 
+    public RedirectedStatus checkForRedirectedRepo(GitHubRepo repo) throws MalformedURLException, HygieiaException {
+        GitHubParsed gitHubParsed = new GitHubParsed(repo.getRepoUrl());
+        String query = gitHubParsed.getBaseApiUrl() + "repos/" + gitHubParsed.getOrgName() + "/" + gitHubParsed.getRepoName();
+
+        ResponseEntity<String> response = makeRestCallGet(query);
+
+        JSONObject queryJSONBody = (JSONObject) parseAsObject(response);
+        String repoUrl = (String) queryJSONBody.get("html_url");
```
Sbrenthughes: can you use the str() method here?

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
```diff
@@ -1042,6 +1056,42 @@ private static DateTime getDate(DateTime dateInstance, int offsetDays, int offse
...
+
+                if (redirectedStatus.isRedirected()) {
+                    String redirectedUrl = redirectedStatus.getRedirectedUrl();
+                    LOG.debug("Repo was redirected from: " + repo.getRepoUrl() + " to " + redirectedUrl);
+                    repo.setRepoUrl(redirectedUrl);
+                    GitHubParsed gitHubParsed = new GitHubParsed(redirectedUrl);
+
+                    JSONParser parser = new JSONParser();
+                    try {
+                        JSONObject variableJSON = (JSONObject) parser.parse((String) query.get("variables"));
```
Sbrenthughes: Consider using the str( method to get the value of `variables `

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
andrewalvintran: Yup, they all redirect to the same repo, which is org/repo3

collectors/scm/github-graphql/src/main/java/com/capitalone/dashboard/collector/DefaultGitHubClient.java
benj58xu: @Sbrenthughes As to redirection, I do not know much about the practices of our interfacing systems, but there are multiple usages in general. For instance, if there are multiple micro services serving client requests directly, we can provide a single URL to the clients and let a proxy to dispatch the requests to the appropriate micro services. In this way, when we add/change micro services, the clients are not impacted.

Sbrenthughes: I don't see how this applies to users changing their repo/ org name.

benj58xu: If we only need to deal with that particular use case, I agree. According to HTTP spec, they should use 301 in this case, never 302.

PR #2833, opened 2019-04-11
Fix security issues with target=_blank

andrewalvintran:
Currently we have some issues that were being flagged on eratocode that involve target="_blank".  Here's a summary of the issue: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/


shriver135:
@andrewalvintran Can you update this branch? It is out of date with master

chtompki:
Unless we're linking away from the dashboard itself, the point is moot.

PR #2852, opened 2019-04-24
Test audit api fix

Sbrenthughes:
1) Fixes logic to look at all test capabilities and use the totals from each for TEST_RESULT_AUDIT_OK/ TEST_RESULT_AUDIT_FAIL
2) Clears Test Suites from response because with many test capabilities the response can get massive.


Sbrenthughes:
@nameisaravind can you validate that this change meets the needs?

PR #2950, opened 2019-06-06
Fix NPE breaking admin page config

fidelcoria-aa:
On the first execution of the Jenkins collector there is a NPE that prevents the collector document for Jenkins from being created. This prevents the General Configuration page from showing the Jenkins configuration.

NPE is avoided by initializing the offending fields in the HudsonSettings class.


rvema:
@fidelcoria-aa Can you fix the compilation errors in your PR, The build is failing due to missing imports in HudsonSettings

PR #2990, opened 2019-06-20
Audit API Response with collector Item optional fields

nameisaravind:
This PR is created to enhance audit api response with collector item optional fields, collectively called as auditEntity in response.


tabladrum:
@nireeshT please see this. If you're ok, merge it.

PR #2994, opened 2019-06-21
The kernel dependency update is configured according to commit # 90 of the core

gonchalo620:
https://github.com/Hygieia/hygieia-core/pull/90

### DIFF
--- a/collectors/scm/github/pom.xml
+++ b/collectors/scm/github/pom.xml
@@ -95,7 +95,7 @@
     <dependency>
       <groupId>com.capitalone.dashboard</groupId>
       <artifactId>core</artifactId>
-      <<version>3.1.2-SNAPSHOT</version>
+      <version>3.1.2-SNAPSHOT</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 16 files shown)


gonchalo620:
> The version needs to be updated only if there is a dependent change on core. Please explain the need for this change

The possibility of connecting ssl with the mongo db in the core project was added.

https://github.com/Hygieia/hygieia-core/commit/4a29517caf95a31e46f554c51dd56f827aec9263

collectors/scm/github/pom.xml
```diff
@@ -95,7 +95,7 @@
     <dependency>
       <groupId>com.capitalone.dashboard</groupId>
       <artifactId>core</artifactId>
-      <version>3.0.4</version>
+      <<version>3.1.2-SNAPSHOT</version>
```
rvema: There is a extra `<` in this pom.xml which is causing the builds to fail. Can you fix it ASAP? I need to merge this and then work on moving this project to its own repository.

PR #3013, opened 2019-06-27
Fix github score category name

### LINKED ISSUE #3011: Repo widget Score has misleading category name
The Score widget says "Commits per day".. but this is just wrong. The calculation is actually "Days with commits". Should The category name be changed or should the category calculation be updated?

I can make a PR with the category name change.

fidelcoria-aa:
you can see at GithubScmWidgetScore.java method getPercentCoverageForDays that the score is actually Days with commits.
fix #3011


fidelcoria-aa:
This actually depends on https://github.com/Hygieia/hygieia-core/pull/94
that one should be merged first to make this build successfully

fidelcoria-aa:
By the way it looks like this:
image
and there is only one day with commits on the graph. 1/14 * 5 days = 0.357 ~ 0.4 so the math checks out

Issue #2979, opened 2019-06-18
Jira Collector NPE, different from 2943


jarbil:
I updated the Jira source last week, so I should have the latest changes. It looks like the docs haven't updated in the past month though. The removed settings just ignored, correct?

Here are the settings I have:
```
#DB
dbname=hdb04
dbusername=huser
dbpassword=****
dbhost=localhost
dbport=27017

# PageSize - Expand contract this value depending on Jira implementation's
# default server timeout setting (You will likely receive a SocketTimeoutException)
feature.pageSize=100

# Delta change date that modulates the collector item task
# Occasionally, these values should be modified if database size is a concern
feature.deltaStartDate=2017-03-01T00:00:00.000000
feature.masterStartDate=2017-03-01T00:00:00.000000
feature.deltaCollectorItemStartDate=2017-03-01T00:00:00.000000
# Chron schedule: S M D M Y [Day of the Week]
feature.cron=0 2/3 * * * *

# ST Query File Details - Required, but DO NOT MODIFY
feature.queryFolder=jiraapi-queries
feature.storyQuery=story
feature.epicQuery=epic

# JIRA CONNECTION DETAILS:
# Enterprise Proxy - ONLY INCLUDE IF YOU HAVE A PROXY
#feature.jiraProxyUrl=
#feature.jiraProxyPort=
feature.jiraBaseUrl=https://jira:8443/
feature.jiraQueryEndpoint=rest/api/2/
# For basic authentication, requires username:password as string in base64
# This command will make this for you:  echo -n username:password | base64
#feature.jiraCredentials=****
feature.jiraCredentials=****

# OAuth is not fully implemented; please blank-out the OAuth values:
feature.jiraOauthAuthtoken=
feature.jiraOauthRefreshtoken=
feature.jiraOauthRedirecturi=
feature.jiraOauthExpiretime=

#############################################################################
# In Jira, general IssueType IDs are associated to various "issue"
# attributes. However, there is one attribute which this collector's
# queries rely on that change between different instantiations of Jira.
# Please provide a string name reference to your instance's IssueType for
# the lowest level of Issues (e.g., "user story") specific to your Jira
# instance.  Note:  You can retrieve your instance's IssueType Name
# listings via the following URI:  https://[your-jira-domain-name]/rest/api/2/issuetype/
#############################################################################
feature.jiraIssueTypeNames=Story

#############################################################################
# In Jira, your instance will have its own custom field created for "sprint" or "timebox" details,
[...]

PR #3036, opened 2019-07-08
Inform user if Epic issue type is not configured

edtan:
This PR is intended to help out with issue #2979:

* Adds the default issue types for a fresh Jira Server install as the default issue types
* Logs an error if the Epic type is not configured

I ran into the NullPointerException seen in issue #2979.  Hopefully these defaults and log message will help operators in the future troubleshoot.

Update: [CLA](https://hygieia.github.io/Hygieia/contribute.html) has been signed.


Sbrenthughes:
@edtan @rvema would it be better to just check the string and add it if it doesn't exist instead of prompting the user to add it?

edtan:
Oh yeah, that makes sense!  For some reason, I was thinking that the Epic issue type was configurable.  But since it is expected to be `Epic`, adding it if it doesn't exist makes sense.  I'll update the PR shortly.

Sbrenthughes:
@rvema Can you provide a quick look and then merge if good?

PR #3053, opened 2019-07-15
dashboardUrl is code derived based on instanceUrl, so remove to reduce confusion; updated pull request to include new parameter for github allowing one to use the projectId instead of the url path.

andy-scollard:


rvema:
@andy-scollard The PR looks good and I will get this merged today.

Issue #3024, opened 2019-07-02
Unable to start bitbucket collector ERROR o.s.s.s.TaskUtils$LoggingErrorHandler - Unexpected error occurred in scheduled task. java.lang.ClassCastException: java.util.Date cannot be cast to java.lang.Long


chenthilan:
It was working well before and might be an issue with the latest code update. I updated my repository today and i get this error when i start the bitbucket jar.

PR #3057, opened 2019-07-17
Handle instances where "lastUpdateTime" can be a Date

cgorshing:
See #3024 for more info.

The `lastUpdateTime` can be a Date in mongo, so handle this case.

Cleaned up the test a bit, added ToStringBuilder (it's already on the classpath). Needed to add a "TestAdapter/Wrapper" so we could get access to the `getOptions()` so the field could be set with the Date value.


rvema:
@cgorshing Thanks for submitting the PR. Have you signed the CLA already? If not please sign the CLA in order for this change to be accepted. https://hygieia.github.io/Hygieia/contribute.html

cgorshing:
@rvema Yes I did last week.

rvema:
Thanks @cgorshing , I will merge this change today!

PR #3118, opened 2019-08-20
Revert "Get Karma unit tests working"

rvema:
Reverts Hygieia/Hygieia#3073 as it broke the widgetmanager


Sbrenthughes:
@edtan we were not able to create widgets with that PR. can you take a look at the code make the changes to fix it and then resubmit a PR?

Issue #3149, opened 2019-09-19
Dashboard widget do not stop collecting when they are destroyed


stevegal:
When you close a dashboard the widgets on that dashboard will continue to collect. By hooking the widget.js to $destroy broadcast event we can easily fix this. 

See PR #3142 for fix

stevegal:
the is fixed by PR #3160

PR #3160, opened 2019-10-02
ensure widgets stop collecting on destroy rather 

stevegal:
this is to address issue #3149 as a separate PR in case the referenced one is too large


rvema:
@Sbrenthughes and @nireeshT Can you also take a look at this PR? minor change and it looks good to me

Issue #3144, opened 2019-09-13
the `fonts` section of the gulp build takes a long time


stevegal:
going through all the `node_modules` folders looking for fonts takes a long time. Would it be easy to just go through the modules whose fonts you want rather than bundle the lot?

rvema:
very good thought, If you have a way around please implement the same and contribute back!

stevegal:
should be closed by #3161

PR #3142, opened 2019-09-11
WIP: Feature/multiple widgets

stevegal:
Note I've not yet re-validated this against the new repo. This is in progress, but I wanted to get some early feedback on this.

PR #3161, opened 2019-10-02
only create fonts from awesome-font

stevegal:
so we're not searching all node modules for all fonts. This really increases build speed.

Raised this in isolation to #3142 in case that PR is too large.

Address: #3144


rvema:
@Sbrenthughes  and @nireeshT Can you also take a look at this PR?

PR #3182, opened 2019-10-31
Fix pull request count for repo widget

shishirchaturvedi:
Fix pull request count for a day, seven day and fourteen days. A pull request count should be aggregated.
resolves : https://github.com/Hygieia/Hygieia/issues/3181

The logic for commits count in repo widget is correct but it is incorrect for PR and issues -
https://github.com/Hygieia/Hygieia/blob/8f14d527dfba42fae5abe65a17224754b887d4e6/UI/src/components/widgets/repo/view.js#L184-L207

If an issue is included in 7 days count then it should be included in 14 days count as well. So `else if` block should be replaced with `if` block,
https://github.com/Hygieia/Hygieia/blob/master/UI/src/components/widgets/repo/view.js#L297-L312

This PR fixes this issue by changing `else if` to `if` block.

### DIFF
--- a/UI/src/components/widgets/repo/view.js
+++ b/UI/src/components/widgets/repo/view.js
@@ -294,14 +294,14 @@
                         lastDayPullContributors.push(pull.userId);
                     }
                 }
-                else if(pull.timestamp >= sevenDays.getTime()) {
+                if(pull.timestamp >= sevenDays.getTime()) {
                     lastsevenDayPullCount++;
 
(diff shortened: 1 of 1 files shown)


rvema:
@shishirchaturvedi just checking if you have signed the CLA, please confirm the samee

shishirchaturvedi:
@rvema Yes, I did sign CLA.

